PR #7, opened 2024-03-13
Report time taken for update proxy agent command and for reporting update proxy agent command 

shahneerali:


proxy_agent_extension/src/service_main.rs
```diff
@@ -121,7 +123,14 @@ fn monitor_thread() {
                     &mut status,
                     &mut status_state_obj,
                 );
-
+                // Time taken to update proxy agent service
+                telemetry::span::SimpleSpan::write_event(
```
ZhidongPeng: proxy_agent_update_command.write_event

proxy_agent_extension/src/service_main.rs
```diff
@@ -132,6 +141,16 @@ fn monitor_thread() {
             &mut restored_in_error,
         );
 
+        // Time taken to report success for proxy agent service after update
+        if status.status == constants::SUCCESS_STATUS.to_string() {
+            telemetry::span::SimpleSpan::write_event(
```
ZhidongPeng: Do not report event every 15 seconds. I suggest 
let proxy_agent_update_reported :Option<SimpleSpan>, if the ProxyAgent version updated, set the value; when the status reached ultimate state proxy_agent_update_reported.write_event(..) and set it to None.

ZhidongPeng (approved):

PR #13, opened 2024-03-29
cargo deb command to generate the deb package

ZhidongPeng:


proxy_agent/Cargo.toml
```diff
@@ -59,3 +59,21 @@ features = [
   "Win32_System_ProcessStatus",
   "Win32_System_Kernel",
 ]
+
+[package.metadata.deb]
```
bingyuyin, line 63: do you need add cfg(not(windows)) to the section below or package.metadata.deb is implict linux based?

bingyuyin (approved):
approve with suggestions

proxy_agent/Cargo.toml
ZhidongPeng: No need since it only works at Linux build.

PR #32, opened 2024-05-02
unmark GuestProxyAgent service before start it

ZhidongPeng:
After we uninstall the azure-proxy-agent package, our GuestProxyAgent service is marked. If we need to install/start that service it again, we need unmark it first.


proxy_agent/ProxyAgent.Connection.log
```diff
@@ -0,0 +1,71 @@
+129387439586104 2024-05-02T02:37:08.149 [INFO]    Connection:1 - Start to match identity 'test'
+129387439586104 2024-05-02T02:37:08.150 [INFO]    Connection:1 - Matched user name 'test' from identity 'test'
```
shahneerali: should delete connection log, maybe add to git ignore

shahneerali (approved):

setharnold:
Hello, I cannot find `systemctl unmark` in the manpage https://www.freedesktop.org/software/systemd/man/latest/systemctl.html# -- what exactly should this do? Was this perhaps a typo for `unmask`, which *does* exist, but which shouldn't be used directly in maintainer scripts? Or is this an unreleased systemd feature that you're using?

I strongly recommend using the debhelper framework for managing these files, as it handles many subtleties for you https://manpages.debian.org/testing/debhelper/dh_installsystemd.1.en.html

I also strongly recommend putting code reformatting changes in their own commits, so that meaningful code changes can be more readily discovered separate from mechanical changes. As far as I can tell, this is 380 lines of changes, which break down into a few categories:

- changing thirty or so comments into echo'd output
- inserting a bug in the Debian postinst
- some test changes that are hard for me to quickly describe
- hundreds of lines of whitespace changes that appear to do nothing

I suspect the bug might have been easier to see if it were one of thirty lines rather than one of 380 lines. (I say `bug`, but maybe you're just using an unreleased feature of systemd? If so, my apologies.)

Thanks

PR #38, opened 2024-05-06
Detect authroziation rule Id changes at each key status api calls

ZhidongPeng:


proxy_agent/src/key_keeper.rs
```diff
@@ -176,6 +178,21 @@ fn poll_secure_channel_status(
         let state = status.get_secure_channel_state();
 
         unsafe {
```
huiminyan2017, line 178: in a brief glance, I don't think the state parse code should be wrapped in "unsafe" closure. unsafe is only needed when dealing with raw pointers, calling FFI etc. Here we are just parsing state structure (yes, send from external), but I think dealing carefully with try .. catch or match statement should be enough. Wrapping all those code into unsafe disabled the compiler's ability to detect potential issue.

bingyuyin (dismissed):

proxy_agent/src/key_keeper.rs
ZhidongPeng: you are correct, let me update it.

huiminyan2017 (approved):

PR #48, opened 2024-05-16
Update Redirector Map + Add or Remove Policy in Redirector Map at Runtime 

shahneerali:


proxy_agent/src/key_keeper.rs
```diff
@@ -317,14 +317,18 @@ fn poll_secure_channel_status(
...
+            redirector::update_imds_redirect_policy(status.get_imds_mode() != DISABLE_STATE);
+            unsafe{
                 *CURRENT_SECURE_CHANNEL_STATE = state.to_string();
             }
-
             // customer has not enforce the secure channel state
             if state == DISABLE_STATE {
                 let message = helpers::write_startup_event(
-                    "Customer has not enforce the secure channel state.",
+                    "Customer has not enforce the sescure channel state.",
```
ZhidongPeng: your typo?

proxy_agent/src/key_keeper/key.rs
```diff
@@ -553,6 +553,50 @@ impl KeyStatus {
         }
     }
 
+    pub fn get_wire_server_mode(&self) -> String {
```
ZhidongPeng, line 556: add unit tests to invoke these two fn in our existing key_status_v1/2_test.

ZhidongPeng (dismissed):

proxy_agent/src/redirector/linux.rs
```diff
@@ -505,6 +507,99 @@ fn lookup_audit_internal(bpf: &Bpf, source_port: u16) -> std::io::Result<AuditEn
...
+                            };
+                        } else {
+                            let local_ip = match get_local_ip() {
+                                Some(ip) => ip,
+                                None => constants::PROXY_AGENT_IP.to_string(),
+                            };
+                            event_logger::write_event(
+                                event_logger::WARN_LEVEL,
+                                format!(
+                                    "update_redirect_policy_internal with local ip address: {}",
```
ZhidongPeng: please help add dest_ipv4, dest_port and local_port to this message

ZhidongPeng (dismissed):

proxy_agent/src/redirector.rs
```diff
@@ -244,6 +244,28 @@ pub fn get_ebpf_file_path() -> PathBuf {
...
+pub fn update_imds_redirect_policy(redirect: bool) {
+    #[cfg(windows)]
+    {
+        windows::update_imds_redirect_policy(redirect);
+    }
+    #[cfg(not(windows))]
+    {
+        linux::update_imds_redirect_policy(redirect);
+    }
+}
```
jeremycline: Another way to do this without writing a wrapper function would be to do something like this:

```suggestion
#[cfg(windows)]
pub use windows::update_imds_redirect_policy;
#[cfg(not(windows))]
pub use linux::update_imds_redirect_policy;
```

You should also be able to use the `pub use linux::{func1, func2, ...};` syntax since it looks like there's a lot of these functions here.

proxy_agent/src/redirector/linux.rs
```diff
@@ -505,6 +507,99 @@ fn lookup_audit_internal(bpf: &Bpf, source_port: u16) -> std::io::Result<AuditEn
...
+pub fn update_imds_redirect_policy(redirect: bool) {
+    update_redirect_policy_internal(
+        constants::IMDS_IP_NETWORK_BYTE_ORDER,
+        constants::IMDS_PORT,
+        redirect,
+    );
+}
+
+fn update_redirect_policy_internal(dest_ipv4: u32, dest_port: u16, redirect: bool) {
+    unsafe {
```
jeremycline: I would recommend not wrapping this whole block in `unsafe`.

The convention in Rust projects with unsafe is to use on the smallest, most minimal block possible, and to include a documentation block on why it's safe to do. See, for example, the [usage in the standard library](https://github.com/rust-lang/rust/blob/e8fbd991287f637f95016a71ddc13438415bbe59/library/core/src/fmt/float.rs#L48) - the reason the code within the `unsafe` block is okay should be documented (assumptions made, potential ways those assumptions could be broken, etc).

It's extremely easy to introduce safety issues when inside an `unsafe` block as you're promising the compiler you'll manually check a lot of the things it automatically checks for you, so making them small and isolated is best.

ZhidongPeng (approved):

PR #54, opened 2024-05-22
Replace custom log implementation with tracing

jeremycline:
The tracing library offers a rich set of features and covers all the implemented features of the logging implementation currently in use.

This commit replaces the various wrappers for the logging service with the tracing equivalent. It also configures tracing to emit logs to stderr and to a rolling file (although I recommend we remove this or make it Windows only as systemd will handle logging for us if we write to stdout or stderr).

Tracing makes it easy to emit structured logs if we prefer, integrate with OpenTelemetry, and more.

I'm marking this as a draft because there's some Span type I've not removed, I've not thoroughly tested it, and I've not thoroughly gone through the diff. It's a pretty huge diff, but it's _mostly_ mechanical search-and-replace stuff.


jeremycline:
I think this is at least ready for a review.

The Linux CI fails because I deleted enough code to drop the coverage below 70%.

proxy_agent_setup/src/main.rs
```diff
@@ -24,13 +24,25 @@ const SERVICE_DISPLAY_NAME: &str = "Microsoft Azure Guest Proxy Agent";
...
+    // Configurable via environment variable for now
+    // https://docs.rs/tracing-subscriber/0.3.18/tracing_subscriber/filter/struct.EnvFilter.html#directives
+    let stderr_layer = tracing_subscriber::fmt::layer()
+        .event_format(format)
+        .with_writer(std::io::stderr)
+        .with_filter(tracing_subscriber::EnvFilter::from_env(
+            "GUEST_PROXY_AGENT_LOG",
+        ));
+    let registry = tracing_subscriber::registry().with(stderr_layer);
+    tracing::subscriber::set_global_default(registry).expect("Unable to configure logging!");
```
LittleBoxOfSunshine, line 39: I'm hesitant to panic here. If the issue is permanent it would result in the VM losing the ability to communicate. It will restart but will loop.

proxy_agent_shared/src/telemetry/span.rs
```diff
@@ -64,20 +63,15 @@ impl SimpleSpan {
...
-            event_logger::INFO_LEVEL,
-            elapsed_massage.to_json_string(),
-            method_name,
-            module_name,
-            logger_key,
-        );
-        elapsed_massage.to_string()
+            ElapsedMessage::new(self.get_elapsed_time_in_millisec(), message.to_string())
+                .to_string();
+        tracing::info!(elapsed_massage,);
```
ZhidongPeng, line 73: will this log write to our event and which will send out our kusto table? I would suggest keep calling event_logger::write_event here but when need log to local file use tracing crate api.

proxy_agent_shared/src/rolling_logger.rs
ZhidongPeng, line 1: let's wrap the tracing crate api (setup with file name, and log actions etc) at our shared crate and leave other files not changed as possible.

proxy_agent_setup/src/main.rs
jeremycline: According to tracing's [documentation](https://docs.rs/tracing/0.1.40/tracing/subscriber/fn.set_global_default.html) it only panics if the global default has already been set, which is either programmer error or a library that is misbehaving. In either case I would hope that it is caught by CI before being published.

However, there is a small chance that in the future the API changes and it can fail for some other reason, so this approach does have downsides. If we do stick with a panic here, I do think I should write a clearer message explaining it's due to programmer error.

How would you recommend proceeding if we can't configure a logger and don't panic? I guess we can announce the failure on stderr, but that's also less noticeable to an administrator than a service failing.

proxy_agent_shared/src/telemetry/span.rs
jeremycline: One of the really great things about tracing it is decouples the event from how it's recorded. We can define whatever policies we want when we configure the tracing subscriber in terms of what events are sent where. This includes not just filtering on level, field, or location, but also _where_ they go. You can send them to `stderr`, or use a pre-existing implementation to hook them up to some other system (like [ETW](https://github.com/microsoft/rust_win_etw/tree/main/win_etw_tracing), or [OpelTelemetry](https://crates.io/crates/tracing-opentelemetry), or implement the interface tracing provides to send them elsewhere.

Can you help me understand the various logging interfaces you've implemented? I see this pushes the string to a queue, and then writes it to a file named after the current current time in nanoseconds. I assume somewhere else is reading that and then doing something?

proxy_agent_shared/src/rolling_logger.rs
jeremycline: Is the idea here to incrementally port everything over after stubbing out the existing interface?

jeremycline: Just a heads-up, I'm not going to continue with this PR until I hear back on this since it's not clear to me if the PR would even be accepted.

PR #59, opened 2024-06-10
Init data_vessel model and move key_keeper Key

ZhidongPeng:
- Created a centralized mod data_vessel
- move the first shared data 'Key' to data_vessel
- encapsulate the Key related functions. 
- modify the callers and relative places.

### DIFF
--- a/proxy_agent/src/data_vessel.rs
+++ b/proxy_agent/src/data_vessel.rs
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation
 // SPDX-License-Identifier: MIT
 
-use crate::key_keeper::key::Key;
+use crate::{common::logger, key_keeper::key::Key};
 use std::sync::mpsc::Sender;
 
 pub enum DataAction {
@@ -25,7 +25,7 @@ impl DataVessel {
 
         std::thread::spawn(move || {
             // chached data are defined here
-            let mut cached_key: Key = Key::empty();
+            let mut cached_key: Key = Key::empty(); // start with empyt key
 
             while let Ok(action) = receiver.recv() {
                 match action {
@@ -53,13 +53,20 @@ impl DataVessel {
     }
 }
 
-impl KeyKeeper for DataVessel {
-    fn update_current_key(&self, key: Key) -> bool {
+/// KeyKeeper implementation
+impl DataVessel {
+    pub fn update_current_key(&self, key: Key) -> bool {
         let (response, receiver) = std::sync::mpsc::channel::<bool>();
         let _ = self
             .sender
             .send(DataAction::KeyKeeperSetKey { key, response });
-        receiver.recv().unwrap_or(false)
+        match receiver.recv() {
+            Ok(result) => result,
+            Err(e) => {
+                logger::write_warning(format!("Failed to update current key with error: {e}"));
+                false
+            }
+        }
     }
 
     fn get_current_key(&self) -> Key {
@@ -68,26 +75,25 @@ impl KeyKeeper for DataVessel {
             .sender
             .clone()
             .send(DataAction::KeyKeeperGetKeyValue { response });
-        receiver.recv().unwrap_or(Key::empty())
+        match receiver.recv() {
+            Ok(key) => key,
+            Err(e) => {
+                logger::write_warning(format!("Failed to get current key with error: {e}"));
+                // return empty key if failed to get the key
+                Key::empty()
+            }
+        }
     }
 
-    fn get_current_key_value(&self) -> String {
+    pub fn get_current_key_value(&self) -> String {
         self.get_current_key().key.to_string()
     }
 
-    fn get_current_key_guid(&self) -> String {
+    pub fn get_current_key_guid(&self) -> String {
         self.get_current_key().guid.to_string()
     }
 
-    fn get_current_key_incarnation(&self) -> Option<u32> {
+    pub fn get_current_key_incarnation(&self) -> Option<u32> {
         self.get_current_key().incarnationId
     }
-}
-
-pub trait KeyKeeper {
-    fn update_current_key(&self, key: Key) -> bool;
-    fn get_current_key(&self) -> Key;
-    fn get_current_key_value(&self) -> String;
-    fn get_current_key_guid(&self) -> String;
-    fn get_current_key_incarnation(&self) -> Option<u32>;
 }
--- a/proxy_agent/src/data_vessel.rs
+++ b/proxy_agent/src/data_vessel.rs
@@ -15,60 +15,79 @@ pub enum DataAction {
     },
 }
 
-pub fn start_receiver_async() -> Sender<DataAction> {
-    let (sender, receiver) = std::sync::mpsc::channel::<DataAction>();
+#[derive(Clone, Debug)]
+pub struct DataVessel {
+    sender: Sender<DataAction>,
+}
+impl DataVessel {
+    pub fn start_new_async() -> Self {
+        let (sender, receiver) = std::sync::mpsc::channel::<DataAction>();
 
-    std::thread::spawn(move || {
-        // chached data are defined here
-        let mut cached_key: Key = Key::empty();
+        std::thread::spawn(move || {
+            // chached data are defined here
+            let mut cached_key: Key = Key::empty();
 
-        while let Ok(action) = receiver.recv() {
-            match action {
-                DataAction::Stop => {
-                    break;
-                }
-                DataAction::KeyKeeperSetKey { key, response } => {
-                    // Set the key from the key keeper
-                    cached_key = key.clone();
-                    let _ = response.send(true);
-                }
-                DataAction::KeyKeeperGetKeyValue { response } => {
-                    // Get the key from the key keeper
-                    let _ = response.send(cached_key.clone());
+            while let Ok(action) = receiver.recv() {
+                match action {
+                    DataAction::Stop => {
+                        break;
+                    }
+                    DataAction::KeyKeeperSetKey { key, response } => {
+                        // Set the key from the key keeper
+                        cached_key = key.clone();
+                        _ = response.send(true);
+                    }
+                    DataAction::KeyKeeperGetKeyValue { response } => {
+                        // Get the key from the key keeper
+                        _ = response.send(cached_key.clone());
+                    }
                 }
             }
-        }
-    });
+        });
 
-    sender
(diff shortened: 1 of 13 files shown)


proxy_agent/src/data_vessel.rs
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    pub fn get_current_key_guid(sender: Sender<DataAction>) -> String {
+        get_current_key(sender).guid.to_string()
+    }
+
+    pub fn get_current_key_incarnation(sender: Sender<DataAction>) -> Option<u32> {
+        get_current_key(sender).incarnationId
+    }
+}
```
jeremycline: I'd recommend replacing this with an `impl` on a struct holding the sender.

With the current approach, this interface isn't discoverable from code that has a `Sender<DataAction>` - you have to just know they exist. New developers likely won't find them and will end up re-implementing these functions in some way given the sender. If you have them as an impl on a struct, however, users have a much easier time discovering them. In fact, since the sender should be a private field, users don't even need to know about the implementation (the sender) and it makes it much easier to replace (which you'll likely want to do if/when you switch to an asynchronous framework like tokio).

proxy_agent/src/data_vessel.rs
```diff
@@ -0,0 +1,93 @@
...
+pub struct DataVessel {
+    sender: Sender<DataAction>,
+}
+impl DataVessel {
+    pub fn start_new_async() -> Self {
+        let (sender, receiver) = std::sync::mpsc::channel::<DataAction>();
+
+        std::thread::spawn(move || {
+            // chached data are defined here
+            let mut cached_key: Key = Key::empty();
```
jeremycline: The more "Rust-y" way to handle what this `empty()` function seems to do is either:

- Implement `Default` for `Key` if what default produces is a useful value.
- Wrap `Key` in an `Option`

I don't know this code so I can't provide a great recommendation on the right approach here, but it seems like what's produced by `Key::empty()` isn't safe or correct to use so I'd recommend the `Option` approach based on that. Of course, you'll need to consider how to handle a request for the key when it's not been set (likely an error?)

proxy_agent/src/data_vessel.rs
```diff
@@ -0,0 +1,93 @@
...
+                        _ = response.send(cached_key.clone());
+                    }
+                }
+            }
+        });
+
+        DataVessel { sender }
+    }
+
+    pub fn stop(&self) {
```
jeremycline, line 51: It might be better to consume `self` since there's no way to use this object after calling `stop`:
```suggestion
    pub fn stop(self) {
```

Obviously since this structure is clone-able this doesn't stop other users from trying to use it after its stopped, but it is a good signal that's the intention.

proxy_agent/src/data_vessel.rs
```diff
@@ -0,0 +1,93 @@
...
+    fn get_current_key_guid(&self) -> String {
+        self.get_current_key().guid.to_string()
+    }
+
+    fn get_current_key_incarnation(&self) -> Option<u32> {
+        self.get_current_key().incarnationId
+    }
+}
+
+pub trait KeyKeeper {
```
jeremycline: I don't see the reason for making this a trait. The only implementation looks to be for `DataVessel`. I would recommend simply implementing these functions on `DataVessel` directly.

proxy_agent/src/data_vessel.rs
```diff
@@ -0,0 +1,93 @@
...
+        DataVessel { sender }
+    }
+
+    pub fn stop(&self) {
+        let _ = self.sender.send(DataAction::Stop);
+    }
+}
+
+impl KeyKeeper for DataVessel {
+    fn update_current_key(&self, key: Key) -> bool {
```
jeremycline: I would not recommend using `bool` to convey errors. Rust uses the `Result` type to indicate fallible calls.

proxy_agent/src/data_vessel.rs
```diff
@@ -0,0 +1,93 @@
...
+        receiver.recv().unwrap_or(false)
+    }
+
+    fn get_current_key(&self) -> Key {
+        let (response, receiver) = std::sync::mpsc::channel::<Key>();
+        let _ = self
+            .sender
+            .clone()
+            .send(DataAction::KeyKeeperGetKeyValue { response });
+        receiver.recv().unwrap_or(Key::empty())
```
jeremycline: This seems dangerous. Quietly returning the default key which seems to be used for cryptographic purposes will likely lead to Bad Things. This `unwrap` will definitely happen when one thread stops this actor and another thread holds this handle and calls this function.

proxy_agent/src/data_vessel.rs
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    fn get_current_key_value(&self) -> String {
+        self.get_current_key().key.to_string()
```
jeremycline: This attribute appears to already be a string anyway, so the call is unnecessary.

proxy_agent/src/data_vessel.rs
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    fn get_current_key_guid(&self) -> String {
+        self.get_current_key().guid.to_string()
```
jeremycline: Similar to above, this attribute is already a string. However, based on the default and the name I'm guessing it's a UUID, so I would recommend changing its type to reflect that ([uuid](https://crates.io/crates/uuid) is a good crate for this and provides everything you'll need.

proxy_agent/src/data_vessel.rs
ZhidongPeng: we need use to_string or clone to deep copy it or the value will moved from this cached_key.

proxy_agent/src/data_vessel.rs
ZhidongPeng: we do not need care about the format at Guest side, it is a just an identification, string type is good enough.

proxy_agent/src/data_vessel.rs
ZhidongPeng: good suggestion.

jeremycline:
On Tue, Jun 11, 2024, at 4:08 PM, Zhidong Peng wrote:
> 
> 
> ***@***.**** commented on this pull request.
> 
> 
> In proxy_agent/src/data_vessel.rs <https://github.com/Azure/GuestProxyAgent/pull/59#discussion_r1635422745>:
> 
> > +            .sender
> +            .send(DataAction::KeyKeeperSetKey { key, response });
> +        receiver.recv().unwrap_or(false)
> +    }
> +
> +    fn get_current_key(&self) -> Key {
> +        let (response, receiver) = std::sync::mpsc::channel::<Key>();
> +        let _ = self
> +            .sender
> +            .clone()
> +            .send(DataAction::KeyKeeperGetKeyValue { response });
> +        receiver.recv().unwrap_or(Key::empty())
> +    }
> +
> +    fn get_current_key_value(&self) -> String {
> +        self.get_current_key().key.to_string()
> 
> we need use to_string or clone to deep copy it or the value will moved from this cached_key.Message ID: ***@***.***>
> 

This can't be true, the function returns an owned `Key` object.

proxy_agent/src/data_vessel.rs
ZhidongPeng: The empty() provides the empty/default values.

proxy_agent/src/data_vessel.rs
ZhidongPeng: hmm, I did not know unwrap_or() will cause unwrap() to panic the application. I thought it should catch the unwrap error and uses Key::empty() value. Let me use match pattern implement it explicitly.

proxy_agent/src/data_vessel.rs
ZhidongPeng: Make sense.

proxy_agent/src/data_vessel.rs
ZhidongPeng: you are right, no need to clone the value.

proxy_agent/src/data_vessel.rs
ZhidongPeng: you are right, no need to clone the value.

proxy_agent/src/data_vessel.rs
jeremycline: This doesn't seem to have been actually resolved.

proxy_agent/src/data_vessel.rs
jeremycline: Yes, I understand what it's doing, what I'm saying is it's not the idiomatic way to do it. The proper way to provide a default is by implementing the `Default` trait. However, I'm also suggesting you might not want to use the default approach at all.

This interface is easy to misuse. The default key can't possibly be useful, right? What's the failure mode if the default key is used by whatever requested it? I assume the best case scenario is the that something the requester wants to do fails, but it could also be the case it's used in such a way that using a well-known default is a security concern.

In either case, you should ensure in this API that it _cannot_ be misused if at all possible, and that is easy to do by wrapping the key in an Option and not having a default function.

proxy_agent/src/data_vessel.rs
```diff
@@ -0,0 +1,99 @@
...
+        let _ = self
+            .sender
+            .clone()
+            .send(DataAction::KeyKeeperGetKeyValue { response });
+        match receiver.recv() {
+            Ok(key) => key,
+            Err(e) => {
+                logger::write_warning(format!("Failed to get current key with error: {e}"));
+                // return empty key if failed to get the key
+                Key::empty()
```
jeremycline, line 83: You misunderstood me, it's not that the `unwrap_or` approach wouldn't work, it's that making up a key if the actor can't be reached instead of returning an error seems incorrect.

ZhidongPeng: if it could not be reached, it logs and returns an invalid empty key. The caller should able to handle from here.

jeremycline: This should return an error instead of hiding it behind an valid-looking invalid value.

bingyuyin (approved):
Approved with suggestions:
will using 'Arc<Mutex < Ke y> >' as the data vessel is simpler that this regarding the implementation?

proxy_agent/src/data_vessel.rs
ZhidongPeng: good catch, seems like the change got reverted from my local. fixed it now.

PR #62, opened 2024-06-14
move static varables from key_keeper and listener mod to data-vessel

ZhidongPeng:
- move more static variables from key_keeper mod to data_vessel
- move static variables from listener mod to data_vessel
- share vessel instance between start_service and stop_service
- return none if key has not update or not able to fetech from data_vessel.

### DIFF
--- a/proxy_agent/src/common/http/http_request.rs
+++ b/proxy_agent/src/common/http/http_request.rs
@@ -21,8 +21,8 @@ impl HttpRequest {
     pub fn new_proxy_agent_request(
         uri: Url,
         mut request: Request,
-        key_guid: String,
-        key: String,
+        key_guid: Option<String>,
+        key: Option<String>,
     ) -> std::io::Result<Self> {
         //connection:Close
         request.headers.add_header(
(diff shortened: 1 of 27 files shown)


jeremycline:
It looks like this branch needs rebasing since #59 was squashed when it was merged and this appears to be based on the branch that got squashed. Git should drop empty commits by default, so a simple `git rebase main` from this branch should do the right thing.

proxy_agent/src/data_vessel.rs
```diff
@@ -80,20 +233,288 @@ impl DataVessel {
...
                 // return empty key if failed to get the key
-                Key::empty()
+                None
             }
         }
     }
 
-    pub fn get_current_key_value(&self) -> String {
-        self.get_current_key().key
+    pub fn get_current_key_value(&self) -> Option<String> {
```
jeremycline, line 241: Is there any reason to not just return the `Key` rather than the string inside it? Users like `new_proxy_agent_request()` want at least the key value and guid so unpacking the struct here seems like more work for callers.

proxy_agent/src/common/http/http_request.rs
```diff
@@ -46,30 +46,32 @@ impl HttpRequest {
...
-                Ok(data) => logger::write_information(format!(
-                    "Computed the signature with input: {}",
-                    data
-                )),
-                Err(e) => {
-                    logger::write_information(format!(
-                        "Failed convert the input_to_sign to string, error {}",
-                        e
-                    ));
+        if let Some(key) = key {
```
jeremycline, line 49: Should this return an error to the requester if the key isn't available? It seems like this should ultimately be turning into a `503 Service Unavailable` error when proxying requests, for example.

proxy_agent/src/data_vessel.rs
```diff
@@ -4,14 +4,73 @@
...
     KeyKeeperSetKey {
-        key: Key,
+        key: Option<Key>,
         response: std::sync::mpsc::Sender<bool>,
     },
     KeyKeeperGetKeyValue {
-        response: std::sync::mpsc::Sender<Key>,
+        response: std::sync::mpsc::Sender<Option<Key>>,
+    },
+    KeyKeeperSetSecureChannelState {
```
jeremycline, line 18: It seems like this is turning into a grab-bag of service state.

I had suggested the actor pattern since it mapped to what the application was previously doing, but if you're going with this approach I think you're better off making a structure you wrap in an `Arc<Mutex<T>>` and pass to all your threads like you're passing the actor handle around.

proxy_agent/src/data_vessel.rs
```diff
@@ -80,20 +233,288 @@ impl DataVessel {
...
+                None
             }
         }
     }
 
-    pub fn get_current_key_value(&self) -> String {
-        self.get_current_key().key
+    pub fn get_current_key_value(&self) -> Option<String> {
+        match self.get_current_key() {
+            Some(key) => Some(key.key),
```
jeremycline, line 243: As I noted above I think you're better off just returning the entire `Key`, but just a head's up for the future when you want to do something like this, you can use the `map` function on the `Option`:
```
self.get_current_key().map(|key| key.key)
```
It's equivalent, but a bit clearer about what's going on than a generic match.

proxy_agent/src/data_vessel.rs
```diff
@@ -80,20 +233,288 @@ impl DataVessel {
...
+        let _ = self
+            .sender
+            .send(DataAction::KeyKeeperSetSecureChannelState { state, response });
+        match receiver.recv() {
+            Ok(result) => result,
+            Err(e) => {
+                logger::write_warning(format!(
+                    "Failed to update secure channel state with error: {e}"
+                ));
+                false
```
jeremycline, line 270: Instead of using the `bool` type to communicate success or failure, it's more idiomatic to use a `Result` and returning an `Error` if something goes wrong. This note applies to lots of functions here so I won't repeat it elsewhere.

jeremycline (changes requested):
I didn't do a thorough review, but I'd recommend that if you just want a single "thing" you can pass around to store program state starting with a struct that you wrap in an `Arc<Mutex<YourStateType>>` will serve you better.

Spawning a thread to just store all the program state does work, but it's surprising and harder.

proxy_agent/src/common/http/http_request.rs
ZhidongPeng: no, it is by design, if the key is not available yet, it will try send the request without signature.

proxy_agent/src/data_vessel.rs
ZhidongPeng: this function is consumed at sender side, it already handled the error. I do not think it needs extra error handling at each caller.

bingyuyin (approved):

PR #64, opened 2024-06-20
Move static variables from key_keeper and proxy_listener mod to Share…

ZhidongPeng:
- Create SharedState strut to hold the static variables from key_keeper and proxy_listener mode 
- Create SharedState instance at the program starts and use Arc::Mutex to clone and share among different threads


bingyuyin (dismissed):

proxy_agent/src/shared_state.rs
```diff
@@ -0,0 +1,157 @@
...
+    proxy_listner_shutdown: bool,
+    connection_count: u128,
+    proxy_listner_status_message: String,
+    // Add more state fields as needed,
+    // keep the fields related to the same module together
+    // keep the fields as private to avoid the direct access from outside via Arc<Mutex<SharedState>>.lock().unwrap()
+    // use wrapper functions to access the state fields, it does quick release the lock
+}
+
+pub fn new_shared_state() -> Arc<Mutex<SharedState>> {
```
jeremycline: It's idiomatic to implement this on the structure:

```
impl SharedState {
    pub fn new() -> Arc<Mutex<SharedState>> {
        Arc::new(Mutex::new(SharedState::default()))
    }
}
```

and then use with `SharedState::new()`

proxy_agent/src/common/http/http_request.rs
```diff
@@ -46,30 +46,32 @@ impl HttpRequest {
...
-                Err(e) => {
-                    logger::write_information(format!(
-                        "Failed convert the input_to_sign to string, error {}",
-                        e
-                    ));
+        if let Some(key) = key {
+            if let Some(key_guid) = key_guid {
+                let input_to_sign = http_request.request.as_sig_input();
+                let authorization_value = format!(
+                    "{} {} {}",
```
hmyan90, line 53: is this a typo: SCHEMA

proxy_agent/src/key_keeper.rs
```diff
@@ -147,72 +138,69 @@ fn poll_secure_channel_status(
...
-                    "Wireserver rule id changed from {} to {}.",
-                    *WIRESERVER_RULE_ID, wireserver_rule_id
-                ));
-                *WIRESERVER_RULE_ID = wireserver_rule_id.to_string();
-                proxy_authentication::set_wireserver_rules(status.get_wireserver_rules());
-            }
+        let imds_rule_id: String = status.get_imds_rule_id();
+        let current_wire_server_rule_id =
+            key_keeper_wrapper::get_wireserver_rule_id(shared_state.clone());
+        if wireserver_rule_id != current_wire_server_rule_id {
```
jeremycline: There's a race condition here where you're reading the current rule, releasing the lock, comparing it, and then re-acquiring the lock and updating it based on the comparison. Another thread could potentially set it such that `wireserver_rule_id != current_wire_server_rule_id` after you read `wireserver_rule_id == current_wire_server_rule_id` and you won't update the rule like you expect.

You can either unconditionally set the value, or (if you want to stick with your wrapper approach) add a wrapper that holds the lock across the read and conditional write.

proxy_agent/src/key_keeper.rs
```diff
@@ -147,72 +138,69 @@ fn poll_secure_channel_status(
...
-            if imds_rule_id != *IMDS_RULE_ID {
-                logger::write_warning(format!(
-                    "IMDS rule id changed from {} to {}.",
-                    *IMDS_RULE_ID, imds_rule_id
-                ));
-                *IMDS_RULE_ID = imds_rule_id.to_string();
-                proxy_authentication::set_imds_rules(status.get_imds_rules());
-            }
+
+        let current_imds_rule_id = key_keeper_wrapper::get_imds_rule_id(shared_state.clone());
```
jeremycline: Similar to above, this introduces a race condition since the lock isn't held between reading the value and conditionally setting it.

jeremycline (commented):
This is definitely an improvement over the unsynchronized globals. If you decide to stick with wrapping functions you'll need to be mindful about TOCTOU-type bugs.

proxy_agent/src/common/http/http_request.rs
```diff
@@ -46,30 +46,32 @@ impl HttpRequest {
...
-                    "Computed the signature with input: {}",
-                    data
-                )),
-                Err(e) => {
-                    logger::write_information(format!(
-                        "Failed convert the input_to_sign to string, error {}",
-                        e
-                    ));
+        if let Some(key) = key {
+            if let Some(key_guid) = key_guid {
```
jeremycline: You can combine this into a single line to remove a bit of nesting here:


```suggestion
        if let (Some(key), Some(key_guid)) = (key, key_guid) {
```

proxy_agent/src/shared_state.rs
```diff
@@ -0,0 +1,222 @@
...
+    pub fn get_current_key_incarnation(shared_state: Arc<Mutex<SharedState>>) -> Option<u32> {
+        get_key(shared_state).map(|k| k.incarnationId)?
+    }
+
+    /// Update the current secure channel state
+    /// # Arguments
+    /// * `shared_state` - Arc<Mutex<SharedState>>
+    /// * `state` - String
+    /// # Returns
+    /// * `bool` - true if the state is update successfully
```
jeremycline, line 83: This doesn't seem to line up with the function is actually doing: it's returning `false` if the current state is the same as the one provided.

proxy_agent/src/shared_state.rs
```diff
@@ -0,0 +1,222 @@
...
+            .current_secure_channel_state
+            .to_string()
+    }
+
+    /// Update the WireServer rule ID
+    /// # Arguments
+    /// * `shared_state` - Arc<Mutex<SharedState>>
+    /// * `rule_id` - String
+    /// # Returns
+    /// * `bool` - true if the rule ID is update successfully
```
jeremycline, line 111: Similar to above, the comment implies this is a fallible call and it's actually an indication of whether something changed or not.

proxy_agent/src/shared_state.rs
```diff
@@ -0,0 +1,222 @@
...
+    pub fn increase_connection_count(shared_state: Arc<Mutex<SharedState>>) -> u128 {
+        let mut state = shared_state.lock().unwrap();
+        let mut connection_count = state.connection_count;
+        if connection_count == u128::MAX {
+            // reset connection count
+            connection_count = 0;
+        }
+        connection_count += 1;
+        state.connection_count = connection_count;
+        connection_count
```
jeremycline: FYI there's [overflowing_add](https://doc.rust-lang.org/std/primitive.u128.html#method.overflowing_add) on primitive integer types to express you want that behavior, so you could write this as 
```suggestion
        (state.connection_count, _) = state.connection_count.overflowing_add(1);
        state.connection_count
```

bingyuyin (approved):

PR #74, opened 2024-07-03
update ebpf version to 0.17.1

ZhidongPeng:
- Update windows ebpf version to 0.17.1
- use ebpf_program_attach function

### DIFF
--- a/proxy_agent/src/redirector/windows/bpf_prog.rs
+++ b/proxy_agent/src/redirector/windows/bpf_prog.rs
@@ -9,6 +9,7 @@ use proxy_agent_shared::misc_helpers;
 use std::env;
 use std::ffi::c_void;
 use std::io::{Error, ErrorKind};
+use std::mem::size_of_val;
 use std::path::PathBuf;
 
 const EBPF_API_FILE_NAME: &str = "EbpfApi.dll";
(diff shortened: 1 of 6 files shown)


proxy_agent/src/redirector/windows/bpf_prog.rs
```diff
@@ -124,20 +130,78 @@ pub fn attach_bpf_prog() -> i32 {
...
+                                logger::write_information("Success disconnected link.".to_string());
+
+                                match bpf_link_destroy(*link) {
+                                    Ok(r) => {
+                                        if r != 0 {
+                                            logger::write_error(format!(
+                                                "Failed to destroy link with error code: {}.",
+                                                r
+                                            ));
+                                            return EBPF_ATTACH_PROGRAM_ERROR;
```
bingyuyin, line 167: may I know what's the link used for?
in case of attach suceed, and link disconnect or destroy failed, do we need to return attach succeed and log error, or just take the whole attach as failed?

ZhidongPeng: This link is internal from ebpf, we do not need it but for best practices to destroy it. 
According to Windows ebpf team, the original call is a wrap 3p ebpf call and they do destroy, hence we best to do the same.

bingyuyin (approved):

PR #78, opened 2024-07-10
Move all static mut variables to stared state

ZhidongPeng:
* move static variables for provision, redirector

* make linux build pass

* move windows ebpf object

* move linux ebpf to shared_state

* move from event_loger

* move extension static mut to ServiceState.


proxy_agent/src/proxy/proxy_summary.rs
```diff
@@ -33,4 +34,17 @@ impl ProxySummary {
             self.responseStatus
         )
     }
+
+    pub fn to_proxy_connection_summary(&self) -> ProxyConnectionSummary {
```
jeremycline: Just a heads up, the more idiomatic way to do this is to implement the [From](https://doc.rust-lang.org/std/convert/trait.From.html) trait from the standard library. This will give you the `from` and `into` functions to convert between the types ([example](https://doc.rust-lang.org/rust-by-example/conversion/from_into.html#from))

proxy_agent/src/redirector/windows/bpf_api.rs
```diff
@@ -4,43 +4,78 @@
...
+static EBPF_API: Lazy<Option<Library>> = Lazy::new(init_ebpf_lib);
+const EBPF_API_FILE_NAME: &str = "EbpfApi.dll";
 
-pub fn load_ebpf_api(bpf_api_file_path: PathBuf) -> std::io::Result<()> {
-    unsafe {
-        match Library::new(bpf_api_file_path.as_path()) {
-            Ok(api) => {
-                *EBPF_API = Some(api);
+pub fn ebpf_api_is_loaded() -> bool {
+    EBPF_API.is_some()
```
bingyuyin, line 20: will the 'is_some()' trigger the 'init_ebpf_lib' function in the Lazy EBPF_API?

bingyuyin (dismissed):

proxy_agent/src/redirector.rs
```diff
@@ -16,6 +16,14 @@ use std::path::PathBuf;
 use std::sync::{Arc, Mutex};
 use std::thread;
 
+#[cfg(not(windows))]
+pub use linux::BpfObject;
+#[cfg(windows)]
+pub use windows::BpfObject;
+/// This is a workaround to allow BpfObject to be sent between threads.
+pub struct RedirectorObject(pub BpfObject);
+unsafe impl Send for RedirectorObject {}
```
jeremycline: By implementing this you're declaring to the compiler "I know this object is safe to send to another thread". Are you certain that `BpfObject` does not rely on thread-local interfaces?  It looks like on Linux this is an alias of `aya::Bpf` which is marked as `Send` so that's fine, but for the Windows version is that documented somewhere, or otherwise demonstrable?

If it _is_ safe to send to another thread you should not be wrapping it in a new type to mark it as Send here. Implement it on the Windows type you're defining and include a `// Safety: <reasons this is safe here>` comment.

If the object _isn't_ safe to send to another thread (e.g. uses thread-local storage) you can still let other threads access it by implementing an Actor that owns the object and uses messages to let other threads work with it.

proxy_agent/src/redirector/linux.rs
```diff
@@ -592,7 +628,7 @@ fn update_redirect_policy_internal(dest_ipv4: u32, dest_port: u16, redirect: boo
 }
 
 #[cfg(test)]
-#[cfg(feature = "test-with-root")]
+//#[cfg(feature = "test-with-root")]
```
jeremycline: Code should be deleted and not commented out. However, based on the diff I don't think you want to delete this.

proxy_agent/src/redirector/windows/bpf_api.rs
```diff
@@ -4,43 +4,78 @@
...
+        thread::sleep(std::time::Duration::from_millis(10));
     }
 
-    Ok(())
+    // after 5 tries, still can't load ebpf api, return None
+    None
+}
+
+fn load_ebpf_api(bpf_api_file_path: PathBuf) -> std::io::Result<Library> {
+    match unsafe { Library::new(bpf_api_file_path.as_path()) } {
```
jeremycline, line 68: This should include a safety comment explaining why it's safe. The reason this function is marked unsafe is that loading involves running initialization functions which could have any requirements to be safe. What, if any, initialization routines will be run from this library?

jeremycline (changes requested):

proxy_agent/src/redirector/linux.rs
ZhidongPeng: good catch

proxy_agent/src/redirector.rs
ZhidongPeng: Updated with more comments from Windows ebpf Team.

proxy_agent/src/redirector/windows/bpf_api.rs
ZhidongPeng: Added comments from Windows ebpf team.

proxy_agent/src/redirector/windows.rs
```diff
@@ -9,52 +9,63 @@ use crate::common::{self, config, constants, helpers, logger};
...
 
-static mut IS_STARTED: bool = false;
-static mut STATUS_MESSAGE: Lazy<String> =
-    Lazy::new(|| String::from("Redirector has not started yet."));
-static mut LOCAL_PORT: u16 = 0;
+pub struct BpfObject(pub *mut bpf_obj::bpf_object);
+// BpfObject is not Send because it contains a raw pointer.
+// However, it is safe to send BpfObject between threads because no bpf_object specific state that is saved in TLS.
+//  bpf_object is reference to the ebpf object, inlcuding names and handles when load to eBPF services, object itself couldnot be updated.
+//  bpf_object is used to interact with the eBPF program and maps, only the eBPF maps elemetry data could be updated at centralized eBPF Maps.
```
jeremycline: // Safety: bpf_object, which is a reference to an eBPF object, has no dependencies on thread-local storage and can
// safely be sent to another thread. This is not explicitly documented in the Windows eBPF library, but the library does
// document it aims to be source-compatible with libbpf[0]. Note that synchronization is required to share this object
// between threads, and care must be taken when using it as libbpf APIs make use of errno[1], which is thread-local.
//
// [0] https://github.com/microsoft/ebpf-for-windows/tree/Release-v0.17.1#2-does-this-provide-app-compatibility-with-ebpf-programs-written-for-linux
// [1] https://libbpf.readthedocs.io/en/v1.4.5/api.html#error-handling

proxy_agent/src/redirector/windows/bpf_api.rs
```diff
@@ -4,43 +4,83 @@
...
+    // after 5 tries, still can't load ebpf api, return None
+    None
+}
+
+fn load_ebpf_api(bpf_api_file_path: PathBuf) -> std::io::Result<Library> {
+    // load ebpf api library dynamically
+    // For thread local init / cleanup -- it only stores and cleans up the device handle to ebpf core.
+    //      So it should be transparent to the caller, and does not impose any extra conditions.
+    // Considering DllMain (for process attach / detach) will be called only once,
+    //      there should not be thread safety concerns too.
```
jeremycline: Technically, `Library::new` is unsafe when called with a library name and the library search path is modified at the same time. You're not using it that way in this module, but the way you've written it makes it easy to refactor code elsewhere and break the assumption here that the function is never being called in that way.

Since you're hard-coding the library path in the function calling this function, I'd recommend moving that down into this function since it's important to the safety of this function. The comment can then be something like:

```
// Safety: Loading a library on  Windows has the following safety requirements:
// 1. The safety requirements of the library initialization and/or termination routines must be met.
// The eBPF for Windows library does not have any safety requirements for its initialization and de-initialization function[0].
// [0] https://github.com/microsoft/ebpf-for-windows/blob/Release-v0.17.1/ebpfapi/dllmain.cpp
//
// 2. Calling this function is not safe in multi-thread scenarios where a library file name is provided and the library
//     search path is modified.
// We satisfy the this requirement by providing the absolute path to the library.
```

proxy_agent/src/redirector/windows.rs
ZhidongPeng: Thanks for your words, I will update to the comments.

proxy_agent/src/redirector/windows/bpf_api.rs
ZhidongPeng: Thanks for your words, I will update to the comments.

shahneerali (approved):

PR #79, opened 2024-07-11
start listener at the given address with retry logic

ZhidongPeng:


proxy_agent/src/proxy/proxy_listener.rs
```diff
@@ -35,14 +35,42 @@ pub fn start_async(port: u16, pool_size: u16, shared_state: Arc<Mutex<SharedStat
...
+    for _ in 0..5 {
+        let listener = TcpListener::bind(addr);
+        match listener {
+            Ok(l) => {
+                return Ok(l);
+            }
+            Err(e) => match e.kind() {
+                std::io::ErrorKind::AddrInUse => {
+                    let message = format!("Failed bind to '{}' with error 'AddrInUse'.", addr);
+                    logger::write_error(message);
```
bingyuyin: do we need to add a sleep here before continue?

ZhidongPeng: yes, I cut it to somewhere else.

PR #90, opened 2024-07-29
Break Windows CMD for RSM Downgrade Scenario 

shahneerali:


proxy_agent/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "azure-proxy-agent"
-version = "9.9.9"     # always 3-number version
+version = "1.1.0"      # always 3-number version
```
bingyuyin: is this version change required ?

proxy_agent_extension/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "ProxyAgentExt"
-version = "9.9.9"      # always 3-number version
```
bingyuyin: same here, is this required?

proxy_agent/Cargo.toml
shahneerali: yes we need to give RSM a version of 1.1 and higher so the can pull a minor version

bingyuyin (approved):

PR #93, opened 2024-08-01
Covert to async with tokio + hyper crates for http logics

ZhidongPeng:
- Convert multiple threads proxy_listener to async mode proxy_server 
- replace custom http modules with hyper crates
- Limit the normal proxied request body size to 100KB


proxy_agent/Cargo.toml
```diff
@@ -18,6 +18,14 @@ url = "2.3.1"                 # parse url string
...
 hex = "0.4.3"                 # hex encode 
 regex = "1.9.5"               # match process name in cmdline
+hyper = { version = "1", features = ["server", "http1", "client"] }
+tokio = { version = "1", features = ["rt", "rt-multi-thread", "net", "macros"] }
+tokio-util = "0.7.11"
+http-body-util = "0.1"
+hyper-util = { version = "0.1", features = ["tokio"] }
+http = "1.1.0"
+## reqwest doesnot build for x86_64-unknown-linux-musl
+# reqwest = "0.12.5"            # http client
```
vmarcella: Can we just get rid of reqwest entirely from our `Cargo.toml` now that we're including `hyper`?

proxy_agent/src/host_clients/imds_client.rs
```diff
@@ -14,49 +14,27 @@ pub struct ImdsClient {
 }
 
 impl ImdsClient {
-    pub fn new(ip: &str, port: u16, shared_state: Arc<Mutex<SharedState>>) -> Self {
+    pub fn new(ip: String, port: u16, shared_state: Arc<Mutex<SharedState>>) -> Self {
         ImdsClient {
             ip: ip.to_string(),
```
vmarcella: Do we need this `to_string()` call now that ip is being passed in as a string?

proxy_agent/src/host_clients/imds_client.rs
```diff
@@ -14,49 +14,27 @@ pub struct ImdsClient {
...
             port,
             shared_state,
         }
     }
 
-    pub fn get_imds_instance_info(&self) -> std::io::Result<InstanceInfo> {
-        const IMDS_URI: &str = "/metadata/instance?api-version=2018-02-01";
+    pub async fn get_imds_instance_info(&self) -> std::io::Result<InstanceInfo> {
+        const IMDS_URI: &str = "metadata/instance?api-version=2018-02-01";
```
vmarcella: I'm not too sure what the actual impact of declaring this const inside the function, but can we define this constant at the top of the module or within `common/constants.rs`? I'm sure the Rust compiler will optimize this to not be redeclared every function call, but I think because it's const we should just move this outside of the function scope.

proxy_agent/src/host_clients/wire_server_client.rs
```diff
@@ -17,153 +13,96 @@ pub struct WireServerClient {
 }
 
 impl WireServerClient {
-    pub fn new(ip: &str, port: u16, shared_state: Arc<Mutex<SharedState>>) -> Self {
+    pub fn new(ip: String, port: u16, shared_state: Arc<Mutex<SharedState>>) -> Self {
         WireServerClient {
             ip: ip.to_string(),
```
vmarcella: Same as with the IMDS client, do we still need this `ip.to_string()` call now that ip is being passed in as a string?

proxy_agent/src/host_clients/wire_server_client.rs
```diff
@@ -17,153 +13,96 @@ pub struct WireServerClient {
...
+        let request = http::get_request(
+            "POST",
+            &url,
+            &headers,
+            Some(xml_data.as_bytes()),
+            None, // post telemetry data does not require signing
+            None,
+        )?;
+        let response = match http::send_request(
+            &self.ip.to_string(),
```
vmarcella: Same as the comment before with regards to ip already being a string

proxy_agent/src/key_keeper.rs
```diff
@@ -97,18 +98,18 @@ fn poll_secure_channel_status(
         if !first_iteration {
             // skip the sleep for the first loop
 
-            let sleep =
-                if key_keeper_wrapper::get_current_secure_channel_state(shared_state.clone())
+            let sleep = if key_keeper_wrapper::get_current_secure_channel_state(shared_state.clone())
```
vmarcella: Would it be worth breaking out the sleep duration calculation into it's own function, or at the very least breaking the conditions into their own variables to make the comparison itself more descriptive?
```rust
#[inline]
pub fn get_sleep_interval(shared_state: Arc<Mutex<SharedState>>, fallback_interval: Duration) -> Duration {
    let in_unknown_state = key_keeper_wrapper::get_current_secure_channel_state(shared_state.clone()) == UNKOWN_STATE;
    let fallback_interval_is_longer = interval > FREQUENT_PULL_INTERVAL;
    let has_not_timed_out = helpers::get_elapsed_time_in_millisec() < FREQUENT_PULL_TIMEOUT_IN_MILLISECONDS;
    
    if in_unknown_state && fallback_interval_is_longer && has_not_timed_out {
        return FREQUENT_PULL_INTERVAL;
    }
    return interval;
}

File: proxy_agent/src/key_keeper.rs [hunk 7]
```diff
@@ -97,18 +98,18 @@ fn poll_secure_channel_status(
...
-                    interval
-                };
+            {
+                // frequent poll the secure channel status every second for the first 5 minutes
+                // until the secure channel state is known
+                FREQUENT_PULL_INTERVAL
+            } else {
+                interval
+            };
             thread::sleep(sleep);
```
vmarcella: With the new concurrent model, does `thread::sleep()` prevent other things from happening inside the main thread? Should we use tokios builtin [sleep](https://docs.rs/tokio/latest/tokio/time/fn.sleep.html) function instead, or does it not matter here?

proxy_agent/src/key_keeper/key.rs
```diff
@@ -629,112 +627,88 @@ impl Clone for Key {
...
-    }
-
-    let status: KeyStatus = serde_json::from_str(&response.get_body_as_string()?)?;
+    let status: KeyStatus = http::get(&url, &headers, None, None, logger::write_warning).await?;
     status.validate()?;
 
     Ok(status)
 }
 
 // base_url must end with '/'
```
vmarcella, line 659: Given that these functions expect `base_url` to end with '/', can we just add an assertion that checks if it does in the function call?

proxy_agent/src/key_keeper/key.rs
```diff
@@ -629,112 +627,88 @@ impl Clone for Key {
...
-            format!(
-                "Cannot parse the json response body {} with error {}",
-                response_body, e
-            ),
-        )),
-    }
+    http::read_response_body(response).await
 }
 
 // base_url must end with '/'
```
vmarcella, line 699: Same as above

proxy_agent/src/proxy/proxy_connection.rs
```diff
@@ -2,21 +2,41 @@
...
+    pub method: String,
+    pub url: String,
     pub ip: String,
     pub port: u16,
 }
 
+impl ConnectionContext {
+    // try make sure the request could skip the sig
+    // and stream the body to the server directly
+    pub fn need_skip_sig(&self) -> bool {
```
vmarcella: NIT: Maybe we can rename this to `can_skip_sig` or `should_skip_sig`?

proxy_agent/src/proxy/proxy_summary.rs
```diff
@@ -6,6 +6,7 @@ use serde_derive::{Deserialize, Serialize};
 #[derive(Serialize, Deserialize)]
 #[allow(non_snake_case)]
 pub struct ProxySummary {
+    pub id: u128,
```
vmarcella, line 9: Is there any reason that the id for the Proxy Summary is so wide?

proxy_agent/src/redirector/windows.rs
```diff
@@ -175,14 +173,14 @@ pub fn lookup_audit(
     bpf_prog::lookup_bpf_audit_map(source_port, shared_state)
 }
 
-pub fn get_audit_from_redirect_context(tcp_stream: &TcpStream) -> std::io::Result<AuditEntry> {
+pub fn get_audit_from_redirect_context(raw_socket: usize) -> std::io::Result<AuditEntry> {
```
vmarcella: NIT: I'd rename this to `raw_socket_id` just to make it a bit more explicit, but I'm not hard pressed about it since you can infer that from it's `usize` type

proxy_agent/src/shared_state.rs
```diff
@@ -108,13 +117,73 @@ impl Default for SharedState {
...
+            mock_server_ip: None,   // test only
+            mock_server_port: None, // test only
             // service
             #[cfg(windows)]
             service_status_handle: None,
         }
     }
 }
 
+pub mod shared_state_wrapper {
```
vmarcella, line 121: NIT: I think this module is better named `shared_state_helpers`, as a lot of the functions are helpers for properly working on the shared state object or accessing things from it.

proxy_agent/src/service.rs
```diff
@@ -32,22 +32,40 @@ pub fn start_service(shared_state: Arc<Mutex<SharedState>>) {
...
-
-    proxy_listener::start_async(constants::PROXY_AGENT_PORT, 20, shared_state.clone());
+    start_service_async(shared_state.clone());
 
     // TODO:: need start the monitor thread and write proxy agent status to the file
     // monitor::start_async(config::get_monitor_duration());
 }
 
+fn start_service_async(shared_state: Arc<Mutex<SharedState>>) {
+    _ = std::thread::Builder::new().spawn(move || {
```
vmarcella: Instead of building a new thread directly, should we let [tokio](https://tokio.rs/tokio/tutorial/spawning) handle that?

proxy_agent/src/shared_state.rs
```diff
@@ -108,13 +117,73 @@ impl Default for SharedState {
...
+    }
+
+    /// Check the cancellation token and return error if it is cancelled
+    pub fn check_cancellation_token(
+        shared_state: Arc<Mutex<SharedState>>,
+        fn_name: &str,
+    ) -> std::io::Result<()> {
+        if shared_state
+            .lock()
+            .unwrap()
```
vmarcella: Will this cause a panic if the lock cannot be unwrapped/acquired?

proxy_agent/src/telemetry/event_reader.rs
```diff
@@ -55,17 +56,29 @@ pub fn start_async(
     _ = thread::Builder::new()
```
vmarcella: Same as a comment left earlier, should we use tokio to spawn this thread?

vmarcella (changes requested):

proxy_agent/Cargo.toml
ZhidongPeng: Let's mark it why we do not use reqwest at this moment.

proxy_agent/src/host_clients/imds_client.rs
ZhidongPeng: good catch.

proxy_agent/src/key_keeper.rs
ZhidongPeng: good catch

proxy_agent/src/key_keeper.rs
ZhidongPeng: it is the same thing since the if condition already breakout for multiple lines

proxy_agent/src/key_keeper/key.rs
ZhidongPeng: make sense, I use the url.join to generate the url string and caught the parse-errors.

proxy_agent/src/proxy/proxy_summary.rs
ZhidongPeng: yeah, we need id when log the proxy summary, we could use this id to lookup the details log.

proxy_agent/src/proxy/proxy_connection.rs
ZhidongPeng: renamed to should_skip_sig.

proxy_agent/src/shared_state.rs
ZhidongPeng: we have the name pattern with wrapper

proxy_agent/src/shared_state.rs
ZhidongPeng: this is by design; we keep all the static variables in this shared_state and use arc/mutex to guarantee the threads safe. if the unwrap failed, the application got into bad state, it has to be panic.

proxy_agent/src/telemetry/event_reader.rs
ZhidongPeng: I make this function as async.

proxy_agent/src/common/http.rs
```diff
@@ -1,478 +1,394 @@
...
-    server_stream: &TcpStream,
-    mut client_stream: &TcpStream,
-    extra_headers: HashMap<&str, &str>,
-) -> std::io::Result<(Response, usize)> {
-    let mut response_reader = BufReader::new(server_stream);
-
-    let mut response_without_body;
-    match read_response_without_body(&mut response_reader) {
-        Ok(r) => response_without_body = r,
+pub fn get_request(
```
vmarcella: Would it make sense to refactor this to `build_request`? At initial glance I thought this was for creating a `GET` request, but after looking at a few examples it was more clear to me that this was for building requests of any http method.

vmarcella (dismissed):
This iteration looks good to me, thanks for making everything else async! I'll give it an approval and wait for Jeremy to give it a further review.

proxy_agent/Cargo.toml
jeremycline: I don't think leaving it here, commented out, makes sense. Put something in an issue tracker for this if you plan to eventually use it, pointing to some upstream issue for whatever the build problem is. As it stands, I'm left with more questions than answers.

proxy_agent/src/common/http.rs
```diff
@@ -1,478 +1,394 @@
...
-
-    // if request contains expects continue header,
-    // the body will send at next socket data
-    if !request.expect_continue_request() {
-        let content_length = request.headers.get_content_length()?;
-        request.set_body(receive_body_internal(&mut reader, content_length)?);
-    }
-
-    Ok(request)
+pub fn htons(u: u16) -> u16 {
```
jeremycline: This is just a confusing (to those unfamiliar with C) rename of `to_be()` without documentation. I don't see any reason for this to exist.

proxy_agent/src/common/http.rs
```diff
@@ -1,478 +1,394 @@
...
 
-pub fn receive_response_data(stream: &TcpStream) -> std::io::Result<Response> {
-    let mut reader = BufReader::new(stream);
-    let mut response = read_response_without_body(&mut reader)?;
-
-    let content_length = response.headers.get_content_length()?;
-    response.set_body(receive_body_internal(&mut reader, content_length)?);
-
-    Ok(response)
+pub fn ntohs(u: u16) -> u16 {
```
jeremycline: Same note as above

proxy_agent/src/common/http.rs
```diff
@@ -1,478 +1,394 @@
...
+    let (host, port) = host_port_from_uri(uri_str)?;
+    let response = match send_request(&host, port, request, log_fun).await {
+        Ok(r) => r,
+        Err(e) => {
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::Other,
+                format!("Failed to send request to {}: {}", uri_str, e),
+            ))
         }
+    };
```
jeremycline: It's more idiomatic to use `map_err`, but as I've recommended before, you should just use the `anyhow` library.

proxy_agent/src/common/http.rs
```diff
@@ -1,478 +1,394 @@
...
+            Ok(f) => f,
+            Err(e) => {
+                return Err(std::io::Error::new(
+                    std::io::ErrorKind::Other,
+                    format!("Failed to get next frame from response: {}", e),
+                ))
             }
         };
+        if let Some(chunk) = frame.data_ref() {
+            body_string.push_str(&String::from_utf8_lossy(chunk));
```
jeremycline: Is `from_utf8_lossy` truly what you want? The response should include a content type header including the character set in use. If it's not UTF-8 it seems like that should be an error instead of a lossy conversion.

proxy_agent/src/common/http.rs
```diff
@@ -1,478 +1,394 @@
...
-                let received_len = received.len();
-                reader.consume(received_len);
-                data.extend(received.iter().copied());
-            }
-            Err(_e) => {
-                // read timeout, assume no more incoming data in the TcpStream
-                return Ok(data);
+pub async fn read_response_body<T>(
+    mut response: hyper::Response<hyper::body::Incoming>,
+    json_format: bool,
```
jeremycline: Wouldn't it be more correct to inspect the Response `Content-Type` header to determine if it's JSON?

proxy_agent/src/proxy/proxy_server.rs
```diff
@@ -0,0 +1,694 @@
...
+    response
+}
+
+async fn handle_request_with_signature(
+    connection: ConnectionContext,
+    request: Request<hyper::body::Incoming>,
+    shared_state: Arc<Mutex<SharedState>>,
+) -> Result<Response<BoxBody<Bytes, hyper::Error>>, hyper::Error> {
+    let (head, body) = request.into_parts();
+    let whole_body = match body.collect().await {
```
jeremycline, line 520: There doesn't appear to be any limit on the request size. Even though it looks like this happens after the authentication check, accepting arbitrarily large requests is probably not a good idea. There's a bunch of pre-built middleware for this kind of thing in [tower](https://tower-rs.github.io/tower/tower/) which should integrate nicely with Hyper (you can probably also implement authentication as a Tower service, as well).

proxy_agent/src/redirector.rs
```diff
@@ -45,14 +45,14 @@ impl AuditEntry {
 
 const MAX_STATUS_MESSAGE_LENGTH: usize = 1024;
 
-pub fn start_async(local_port: u16, shared_state: Arc<Mutex<SharedState>>) {
-    thread::spawn(move || {
-        start(local_port, shared_state);
+pub async fn start_async(local_port: u16, shared_state: Arc<Mutex<SharedState>>) {
+    tokio::spawn(async move {
+        start(local_port, shared_state).await;
     });
```
jeremycline: This function doesn't need to be marked `async` since it never needs to await anything. Additionally, there's no need to wrap the `start()` function in an async block as it's already returning a future you can pass to `tokio::spawn`:
```suggestion
pub fn start_async(local_port: u16, shared_state: Arc<Mutex<SharedState>>) {
    tokio::spawn(start(local_port, shared_state));
```

This applies to a number of other functions in this PR as well.

proxy_agent/src/shared_state.rs
```diff
@@ -72,6 +77,10 @@ impl SharedState {
 impl Default for SharedState {
     fn default() -> Self {
         SharedState {
+            runtime: Some(Arc::new(Mutex::new(
+                tokio::runtime::Runtime::new().unwrap(),
+            ))),
```
jeremycline: I really wouldn't recommend this. Use the [tokio::main](https://docs.rs/tokio/latest/tokio/attr.main.html) macro.

proxy_agent/src/shared_state.rs
```diff
@@ -108,13 +117,73 @@ impl Default for SharedState {
             telemetry_reader_shutdown: false,
             telemetry_logger_shutdown: false,
             telemetry_logger_status_message: UNKNOWN_STATUS_MESSAGE.to_string(),
+            mock_server_ip: None,   // test only
```
jeremycline: I'd recommend not putting test things here, but if you absolutely have to you can put them behind `#cfg(test)`

proxy_agent/src/telemetry/event_reader.rs
jeremycline: I'd recommend using the `tokio::main` macro which will set up the runtime for the main thread. Then you can get rid of the entire `get_runtime` thing and just spawn the task.

proxy_agent/src/telemetry/event_reader.rs
```diff
@@ -447,8 +457,10 @@ mod tests {
         assert!(THREAD_PRIORITY_VERIFY_DONE.load(Ordering::Relaxed));
     }
 
-    #[test]
-    fn test_event_reader_thread() {
+    // this test is to test the event reader thread, it reads events from the events folder and send to wire server
+    // it requires more threads to run server and client
```
jeremycline: Do the tests hang or something with a single-thread runtime? They shouldn't, and if they do that's an indication that something's wrong with the code.

proxy_agent/src/proxy/proxy_server.rs
```diff
@@ -0,0 +1,694 @@
...
+                continue;
+            }
+        };
+
+        if shared_state_wrapper::get_cancellation_token(shared_state.clone()).is_cancelled() {
+            let message = "Stop signal received, stop the listener.";
+            proxy_listener_wrapper::set_status_message(shared_state.clone(), message.to_string());
+            logger::write_warning(message.to_string());
+            return Ok(());
+        }
```
jeremycline: This is not really how the cancellation token is meant to be used. Instead, you can `select!` on the cancellation token along with the future you want to let someone else cancel.

jeremycline (commented):
I didn't do a full review, but this has some notes with what I had time to get to today.

proxy_agent/src/common/http.rs
ZhidongPeng: This is particular function in networking/socket programming. Let me add bit comments to this function.

jeremycline: Yes, I'm familiar with the C function. There's no reason to wrap a clear, well-documented function provided by the Rust standard library with a different one that only makes sense if you know C. It provides no value, but it does make readers have to do the translation in their head. Please remove it.

proxy_agent/src/common/http.rs
ZhidongPeng: used map_err

proxy_agent/src/common/http.rs
ZhidongPeng: make sense.

proxy_agent/src/proxy/proxy_server.rs
ZhidongPeng: Good idea to keep GPA service not OOM. Let me discuss with the host team to decide the size limitation. Temporally set limit to 100KB.

proxy_agent/src/proxy/proxy_server.rs
ZhidongPeng: I switched to Shutdown state in shared_state as before.

proxy_agent/src/telemetry/event_reader.rs
ZhidongPeng: removed the "flavor = "multi_thread", worker_threads = 2"

proxy_agent/src/shared_state.rs
ZhidongPeng: removed.

proxy_agent/src/shared_state.rs
ZhidongPeng: switched to tokio::main and convert all proxy_agent to async mode.

proxy_agent/src/redirector.rs
ZhidongPeng: removed the "async move" within all tokio::spwan function.

proxy_agent/src/common/http.rs
ZhidongPeng: read content_tye (json/xml), charset_type (utf-8/utf-16) from Content-Type header.

proxy_agent/src/common/http.rs
ZhidongPeng: removed.

proxy_agent/Cargo.toml
ZhidongPeng: removed.

proxy_agent/src/main.rs
```diff
@@ -72,22 +86,47 @@ fn main() {
...
+    // // hence we have to start with normal fn and create a runtime to start async functions
+    // let rt = tokio::runtime::Builder::new_multi_thread()
+    //     .enable_all()
+    //     .build()
+    //     .unwrap();
+    // rt.block_on(async {
+    //     if let Err(e) = windows::run_service().await {
+    //         logger::write_error(format!("{e}"));
+    //     }
+    // });
```
jeremycline: Let's not check in commented-out code.

proxy_agent/src/main.rs
```diff
@@ -72,22 +86,47 @@ fn main() {
...
 
-    if let Err(e) = windows::run_service(args) {
-        logger::write_error(format!("{e}"));
+    match ASYNC_RUNTIME_HANDLE.get() {
+        Some(handle) => {
+            handle.spawn(windows::run_service());
+        }
+        None => {
+            logger::write_error("Failed to get the tokio runtime handle.".to_string());
+        }
```
jeremycline: I don't think it makes sense for the service to continue running in this case. This is a fatal programmer error since this can only happen if `ASYNC_RUNTIME_HANDLE` wasn't initialized prior to this function being invoked. Something like `let handle = ASYNC_RUNTIME_HANDLE.get().expect("You must provide the Tokio runtime handle before this function is called")` or something to that effect.

proxy_agent/src/main.rs
```diff
@@ -72,22 +86,47 @@ fn main() {
...
-fn proxy_agent_windows_service_main(args: Vec<OsString>) {
+fn proxy_agent_windows_service_main(_args: Vec<OsString>) {
     // start the Instant to calculate the elapsed time
     _ = helpers::get_elapsed_time_in_millisec();
 
-    if let Err(e) = windows::run_service(args) {
-        logger::write_error(format!("{e}"));
+    match ASYNC_RUNTIME_HANDLE.get() {
+        Some(handle) => {
+            handle.spawn(windows::run_service());
```
jeremycline: Is the Windows service lifetime tied to `proxy_agent_windows_service_main`? In other words, do you want this call to block until `run_service` completes? If so you want [block_on](https://docs.rs/tokio/latest/tokio/runtime/struct.Handle.html#method.block_on).

proxy_agent/src/monitor.rs
```diff
@@ -8,26 +8,23 @@ use crate::{
...
-    _ = thread::Builder::new()
-        .name("monitor".to_string())
-        .spawn(move || {
-            start(interval, shared_state);
-        });
+pub async fn start_async(interval: Duration, shared_state: Arc<Mutex<SharedState>>) {
+    tokio::spawn(async move {
+        start(interval, shared_state).await;
+    });
 }
```
jeremycline:
```suggestion
pub fn start_async(interval: Duration, shared_state: Arc<Mutex<SharedState>>) {
    tokio::spawn(start(interval, shared_state));
}
```

proxy_agent/src/provision.rs
```diff
@@ -49,7 +49,7 @@ pub fn provision_timeup(provision_dir: Option<PathBuf>, shared_state: Arc<Mutex<
     }
 }
 
-pub fn start_event_threads(shared_state: Arc<Mutex<SharedState>>) {
+pub async fn start_event_threads(shared_state: Arc<Mutex<SharedState>>) {
```
jeremycline: This function doesn't call any asynchronous functions. It doesn't need to be marked async as it never needs to await any results.

proxy_agent/src/proxy/proxy_connection.rs
```diff
@@ -2,21 +2,41 @@
...
+pub struct Connection {}
 
+pub struct ConnectionContext {
+    pub id: u128,
+    pub stream: Arc<Mutex<TcpStream>>,
+    pub client_addr: SocketAddr,
     pub now: Instant,
-    pub cliams: Option<Claims>,
+    pub claims: Option<Claims>,
+    pub method: String,
```
jeremycline: Use the [hyper-provided type for this](https://docs.rs/hyper/latest/hyper/struct.Method.html)

proxy_agent/src/proxy/proxy_server.rs
```diff
@@ -0,0 +1,689 @@
...
+            return Ok(empty_response(StatusCode::MISDIRECTED_REQUEST));
+        }
+    };
+    Connection::write(connection_id, claim_details.to_string());
+    connection.claims = Some(claims.clone());
+
+    // Get the dst ip and port to remote server
+    let (ip, port);
+    ip = redirector::ip_to_string(entry.destination_ipv4);
+    port = u16::from_be(entry.destination_port); // convert a 16-bit number from network byte order to host byte order
```
jeremycline: Why doesn't whatever type `entry` is handle converting the port to the host byte order? This shouldn't be a concern of the user of the type as it's easy to forget or otherwise mess up.

proxy_agent/src/proxy/proxy_server.rs
```diff
@@ -0,0 +1,689 @@
...
+    Connection::write(connection_id, format!("Use lookup value:{ip}:{port}."));
+    connection.ip = ip.to_string();
```
jeremycline: This `ip` variable is coming from a function called `ip_to_string`. I've not got this open in an IDE so I'm not sure what the actual type is, but either the function is mis-named, or this is unnecessary. In any case, I don't think you should store it in this struct as a string, keep it as a meaningful type.

proxy_agent/src/proxy/proxy_server.rs
```diff
@@ -0,0 +1,689 @@
...
+    connection.port = port;
+
+    // authenticate the connection
+    if !proxy_authentication::authenticate(
+        ip.to_string(),
+        port,
+        connection_id,
+        request.uri().to_string(),
```
jeremycline: I don't know how this function is using the URI, but it seems more appropriate to use the `Uri` type you get from the function call instead of turning it into a string.

proxy_agent/src/proxy/proxy_server.rs
```diff
@@ -0,0 +1,689 @@
...
+
+    // Add required headers
+    let host_claims = format!(
+        "{{ \"{}\": \"{}\"}}",
+        constants::CLAIMS_IS_ROOT,
+        claims.runAsElevated
+    );
+    proxy_request.headers_mut().insert(
+        HeaderName::from_static(constants::CLAIMS_HEADER),
+        HeaderValue::from_str(&host_claims).unwrap(),
```
jeremycline: Replace the `unwrap()` here with proper error handling.

proxy_agent/src/proxy/proxy_server.rs
```diff
@@ -0,0 +1,689 @@
...
+    );
+    proxy_request.headers_mut().insert(
+        HeaderName::from_static(constants::DATE_HEADER),
+        HeaderValue::from_str(&misc_helpers::get_date_time_rfc1123_string()).unwrap(),
```
jeremycline: Replace the `unwrap()` here with proper error handling

proxy_agent/src/proxy/proxy_server.rs
```diff
@@ -0,0 +1,689 @@
...
+            connection.id,
+            format!(
+                "The imcoming request body size {} exceeds the limit {}",
+                size, MAX_REQUEST_BODY_SIZE
+            ),
+        );
+        return Ok(empty_response(StatusCode::PAYLOAD_TOO_LARGE));
+    }
+
+    let whole_body = match body.collect().await {
```
jeremycline: What's to stop a client from lying in the `Content-Length` header? Does `collect()` on body take the `Content-Length` header into account in all cases (I suspect the answer is "no" and see no unit tests here to prove otherwise)?

As I previously commented, use a tried and tested implementation to limit body sizes rather than implementing it yourself. [Tower has middleware for this](https://docs.rs/tower-http/latest/tower_http/limit/index.html).

proxy_agent/src/proxy/windows.rs
```diff
@@ -111,9 +103,14 @@ fn load_users() -> HashMap<u64, &'static str> {
...
-pub fn get_user(logon_id: u64) -> (String, Vec<String>) {
+pub fn get_user(
+    shared_state: Arc<Mutex<SharedState>>,
+    logon_id: u64,
+) -> std::io::Result<(String, Vec<String>)> {
+    shared_state_wrapper::check_cancellation_token(shared_state.clone(), "windows::get_user")?;
+
     unsafe {
-        let mut user_name = "undefined".to_string();
+        let mut user_name;
```
jeremycline: This change doesn't seem related to everything else.

As I mentioned elsewhere, this isn't not how to use cancellation tokens; they should be `select!`ed on with whatever future you want to let someone else cancel and the callee doesn't need to have any knowledge of it.

As an aside, this `unsafe` block needs to be replaced with minimal blocks around the actually-unsafe calls. Please do that in a separate PR, however.

proxy_agent/src/service.rs
```diff
@@ -4,20 +4,18 @@
...
 use std::sync::{Arc, Mutex};
 use url::Url;
 
-#[cfg(not(windows))]
-use std::thread;
 #[cfg(not(windows))]
 use std::time::Duration;
 
-pub fn start_service(shared_state: Arc<Mutex<SharedState>>) {
+pub async fn start_service(shared_state: Arc<Mutex<SharedState>>) {
```
jeremycline: As with several other functions in this PR, this function never awaits anything and does not need to be marked as `async.

jeremycline (commented):
Once again, this isn't a complete review, I'm time-boxing this and there's plenty of substantial changes to be made here.

proxy_agent/src/main.rs
ZhidongPeng: good catch, removed.

proxy_agent/src/main.rs
ZhidongPeng: make sense, replaced with your suggestion.

proxy_agent/src/main.rs
ZhidongPeng: we need block_on

proxy_agent/src/monitor.rs
ZhidongPeng: removed this one-line wrapper fun

proxy_agent/src/provision.rs
ZhidongPeng: yes, it is no longer needed after previous changes.

proxy_agent/src/proxy/proxy_connection.rs
ZhidongPeng: switched to hyper::Method struct instead of String.

proxy_agent/src/proxy/proxy_server.rs
ZhidongPeng: This entry is get from c/ebpf directly, I can wrap this property into a function to mitigate this problem.

proxy_agent/src/proxy/proxy_server.rs
ZhidongPeng: redirector::ip_to_string is convert the ip v4 address in u32 format to string, I switched to use IPv4Addr::from_bits fun

proxy_agent/src/proxy/proxy_server.rs
ZhidongPeng: request.uri() is relative Url here, it is hyper::Uri struct, while in our authenticate logic, we used another Url crate struct which support query_pairs(). 
I will convert the request.uri() to Url crate struct and pass down.

proxy_agent/src/proxy/proxy_server.rs
ZhidongPeng: resolved, if failed to insert this header, it will response bad gateway error

proxy_agent/src/proxy/proxy_server.rs
ZhidongPeng: resolved, if failed to insert this header, it will response bad gateway error

proxy_agent/src/proxy/proxy_server.rs
ZhidongPeng: tower-http middleware controls all the requests, while in our case, it skips the collect() if the requests do not require signature. Like if VMAgent put VMAgent zip log file (up to 100MB) to host, it will stream the data to host without any buffer in GPA.

proxy_agent/src/proxy/windows.rs
ZhidongPeng: tokio::select! is designed for asynchronous contexts or async fn, in my case, it is not async fn, I have to use cancellation_token.is_cancelled() to determine this function should move forward or return interrupted error earlier.

ZhidongPeng: removed the large unsafe section, instead of, I added some unsafe to the statements only needed.

proxy_agent/src/service.rs
ZhidongPeng: you are right, the async is no longer needed after I moved to tokio::spawn instead of calling the async methods.

reviewed all the async keywords in the project and removed unnecessary ones.

proxy_agent/src/proxy/proxy_server.rs
ZhidongPeng: Added comments to these fields and functions to get the value properly.

proxy_agent/src/proxy/proxy_server.rs
ZhidongPeng: I also converted the string type to IIv4Addr struct.

proxy_agent/src/proxy/proxy_server.rs
ZhidongPeng: done.

proxy_agent/src/proxy/proxy_server.rs
ZhidongPeng: done.

proxy_agent/src/proxy/windows.rs
jeremycline: Please back out all these changes. Fix the unsafe block in a separate PR as I asked, this one is already too big.

Regarding this function generally: you're using the `windows-sys` API, which is following the normal Rust convention of providing the unsafe binding to the FFI. There's a `windows` crate that provides a safe wrapper around that API. Is there no API in that crate (for example [this one](https://microsoft.github.io/windows-docs-rs/doc/windows/System/struct.User.html#method.GetFromId)) that does what you need? If there is no safe function, perhaps we can contribute one to the crate. 

This implementation is clearly wrong in at least one place: you call `let _status = Identity::LsaGetLogonSessionData(&luid, data.as_mut_ptr());` and then access data without checking the return value, but [it's a fallible call](https://learn.microsoft.com/en-us/windows/win32/api/ntsecapi/nf-ntsecapi-lsagetlogonsessiondata#return-value).

Finally, regarding cancellation: you don't use `select!` in this function, you would use it further up the call stack on a relatively long-running task you may want to cancel. It makes no sense to have a function that accepts a cancellation token, then immediately checks it and returns an error if it's cancelled. Why would the caller even call the function if it's already been cancelled? What's the run time of this function in the worst case? If it's relatively long, you shouldn't be calling it in any asynchronous context. If it's not relatively long, what's the value of cancelling it?

proxy_agent/src/proxy/proxy_server.rs
jeremycline: I am reasonably confident that you can design things such that the middleware is only applied to a sub-set of requests. Axum is a web framework that uses Hyper and Tower for middleware and you can do this. If you can't figure it out I'm happy to look harder at the implementation details, but your current approach is almost certainly incorrect.

To be honest, you might be able to just use Axum directly, so you don't re-invent that particular wheel.

proxy_agent/src/proxy/proxy_server.rs
jeremycline: Remove the `pub` marker from the fields so users can't get a hold of them without going through the accessor function.

proxy_agent/src/proxy/proxy_server.rs
jeremycline: [hyper::Uri](https://docs.rs/hyper/latest/hyper/struct.Uri.html) has scheme, host, and port components as well so your comment about it being relative doesn't make sense. It doesn't handle parsing the query, but `Url` just uses [form_urlencode](https://docs.rs/form_urlencoded/1.2.1/form_urlencoded/fn.parse.html) to provide that feature.

proxy_agent/src/proxy/windows.rs
ZhidongPeng: got it, I will remove the check_cancellation_token here and just keep the select! in the upper caller.

proxy_agent/src/proxy/proxy_server.rs
ZhidongPeng: used use tower service as middleware to limit the request body size

proxy_agent/src/main.rs
```diff
@@ -72,22 +86,31 @@ fn main() {
...
-            _ = service_dispatcher::start(constants::PROXY_AGENT_SERVICE_NAME, ffi_service_main);
+            //
+            let main = std::thread::spawn(|| {
+                _ = service_dispatcher::start(
+                    constants::PROXY_AGENT_SERVICE_NAME,
+                    ffi_service_main,
+                );
+            });
+
+            main.join().unwrap();
```
jeremycline, line 97: Ah, this will block one of the runtime threads which defeats the purpose of spawning the thread.

There's a couple options here, but the easiest is to (ab)use [spawn_block](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html). You're not _really_ meant to use it for things that last the length of the program, but the default pool size is 512 and you're not otherwise doing much heavy i/o that doesn't have async support.

proxy_agent/src/common/http.rs
```diff
@@ -1,478 +1,456 @@
...
+                let charset_type;
+                if content_type_str.contains("utf-8") {
+                    charset_type = "utf-8";
+                } else if content_type_str.contains("utf-16") {
+                    charset_type = "utf-16";
+                } else if content_type_str.contains("utf-32") {
+                    charset_type = "utf-32";
+                } else {
+                    charset_type = "unknown";
+                }
```
jeremycline, line 88: There's a [mime](https://docs.rs/mime/latest/mime/) library you can use to parse things. It's already in your dependency tree via tower and it's very widely used.

jeremycline (commented):
Once again, I've not done a full review, nor do I expect that I will have the time to do so. This pull request is huge and continues to add unrelated changes despite my requests to not do so.

My recommendation would be to focus on cleaning up some of the foundations of this project before focusing on anything else. Get rid of the custom logging and use tracing. Clean up the error handling with `anyhow`. Reviewing this is difficult in part because the error handling and logging is cumbersome.

proxy_agent/src/main.rs
ZhidongPeng: I forget why I use the thread::spawn, it looks like I do not need to spawn and block it here.

proxy_agent/src/common/http.rs
ZhidongPeng: Thanks!

ZhidongPeng:
This big PR break down into 3 smaller ones.

PR #103, opened 2024-08-23
Remove monitor.rs and replace its status with proxy_agent_status

ZhidongPeng:


proxy_agent/src/proxy_agent_status.rs
```diff
@@ -95,7 +94,12 @@ fn proxy_agent_status_new(shared_state: Arc<Mutex<SharedState>>) -> ProxyAgentSt
     ProxyAgentStatus {
         version: misc_helpers::get_current_version(),
         status,
-        monitorStatus: monitor::get_status(shared_state.clone()),
+        // monitorStatus is proxy_agent_status itself status
+        monitorStatus: ProxyAgentDetailStatus {
+            status: ModuleState::RUNNING.to_string(),
```
shahneerali, line 99: if monitor status is proxy agent status would monitorStatus be dependent on let mut status on line 86?

shahneerali (approved):

PR #106, opened 2024-08-24
Fix: get_url_path_and_canonicalized_parameters should not percent-decoded the query pairs.

ZhidongPeng:
handle two more case, duplicate query keys and no query value

### DIFF
--- a/proxy_agent/src/common/http/request.rs
+++ b/proxy_agent/src/common/http/request.rs
@@ -200,15 +200,22 @@ impl Request {
 
         let path = String::from(url.path());
 
-        let parameters = url.query_pairs();
+        // The query_pairs method from the Url struct returns an iterator over the query pairs, which are automatically percent-decoded
+        // To get the raw percent-encoded values, we need to manually parse the query string
+        let query = url.query().unwrap_or("");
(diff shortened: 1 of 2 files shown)


proxy_agent/src/common/http/request.rs
```diff
@@ -200,24 +200,43 @@ impl Request {
...
-        let parameters = url.query_pairs();
-        let mut pairs: HashMap<String, String> = HashMap::new();
-        let mut canonicalized_parameters = String::new();
-        if parameters.count() > 0 {
-            for p in parameters {
-                // Convert the parameter name to lowercase
-                pairs.insert(p.0.to_lowercase(), p.1.to_string());
+        // The query_pairs method from the Url struct returns an iterator over the query pairs, which are automatically percent-decoded
+        // To get the raw percent-encoded values, we need to manually parse the query string
+        let query = url.query().unwrap_or("");
```
shahneerali: instead of using unwrap, can you use match and log if error

ZhidongPeng: query() returns Option<&str> not err. Also, we already logged the Uri when received the request.

shahneerali (approved):

PR #109, opened 2024-08-28
Update --status command to query current provision result

ZhidongPeng:
- Update --status command to query current provision result
- keep failed messages only for provision state
- provision state query request checks for Metadata header
- provision state check request to notify the key_keepr to poll status.

### DIFF
--- a/proxy_agent/src/provision.rs
+++ b/proxy_agent/src/provision.rs
@@ -11,6 +11,7 @@ use proxy_agent_shared::misc_helpers;
 use proxy_agent_shared::telemetry::event_logger;
 use serde_derive::{Deserialize, Serialize};
 use std::io::{Error, ErrorKind};
+use std::net::Ipv4Addr;
 use std::path::PathBuf;
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
@@ -174,7 +175,7 @@ pub fn get_provision_state(shared_state: Arc<Mutex<SharedState>>) -> ProivsionSt
     }
 }
 
-/// Get current provision status and wait until provision finished or timeout
+/// Get current GPA service provision status and wait until provision finished or timeout
 /// it serves for --status --wait command line option
 pub async fn get_provision_status_wait(port: u16, duration: Option<Duration>) -> (bool, String) {
     loop {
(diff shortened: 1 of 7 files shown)


proxy_agent/src/provision.rs
```diff
@@ -129,14 +138,59 @@ fn write_provision_state(
...
+pub async fn get_provision_status_wait(port: u16, duration: Option<Duration>) -> (bool, String) {
     loop {
-        let provision_status = get_provision_status(provision_dir.clone());
-        if provision_status.0 {
-            return provision_status;
+        let provision_state = get_current_provision_status(port);
+        let (finished, message) = match provision_state {
+            Ok(state) => (state.finished, state.errorMessage),
+            Err(e) => {
+                println!(
```
shahneerali, line 185: instead of println should you use logger or raise exception?

proxy_agent/src/provision.rs
```diff
@@ -146,57 +200,96 @@ pub async fn get_provision_status_wait(
             }
         }
 
-        return provision_status;
+        // wait timedout return as 'not finished'
+        return (false, String::new());
```
shahneerali, line 205: should String::new() be "not finished".to_string() ?

proxy_agent/src/provision.rs
```diff
@@ -146,57 +200,96 @@ pub async fn get_provision_status_wait(
...
 //  String - provision error message, empty means provision success or provision failed.
-fn get_provision_status(provision_dir: Option<PathBuf>) -> (bool, String) {
-    let provision_dir = provision_dir.unwrap_or_else(config::get_keys_dir);
-
-    let status_file: PathBuf = provision_dir.join(STATUS_TAG_FILE_NAME);
-    if !status_file.exists() {
-        return (false, String::new());
+fn get_current_provision_status(port: u16) -> std::io::Result<ProivsionState> {
+    let provision_url =
+        url::Url::parse(&format!("http://127.0.0.1:{}{}", port, PROVISION_URL_PATH)).map_err(
```
shahneerali: believe http://127.0.0.1 is in constants already

proxy_agent/src/provision.rs
ZhidongPeng: the first boolean parameter 'false' means not finished.

proxy_agent/src/provision.rs
ZhidongPeng: make sense, I will replace 127.0.0.1 with our constant.

ZhidongPeng: replace it with Ipv4Addr::LOCALHOST

proxy_agent/src/provision.rs
```diff
@@ -146,57 +201,99 @@ pub async fn get_provision_status_wait(
...
+    ))
+    .map_err(|e| {
+        std::io::Error::new(
+            std::io::ErrorKind::Other,
+            format!("Failed to parse provision url with error: {}", e),
+        )
+    })?;
+    let mut req = Request::new(PROVISION_URL_PATH.to_string(), "GET".to_string());
+    req.headers
+        .add_header(constants::METADATA_HEADER.to_string(), "True ".to_string());
```
ghost, line 228: Because I see GPA code various places either adding or checking different headers. Can you wrap into some helper functions, so they are in a central place. (Maybe hyper_client when you move over.)

ghost (approved):

proxy_agent/src/provision.rs
ZhidongPeng: hyper_client has build_request function accepts hashmap for all the headers.

ZhidongPeng: will revisit it in hyper-client PR.

shahneerali (approved):

PR #115, opened 2024-09-13
migrate proxy listener to use hyper and tower crates

ZhidongPeng:
- uses the `hyper` crate to handle the HTTP requests and responses
- uses the `tower` crate to limit the incoming request body size
- remove our own http mod
- remove url crate dependency


proxy_agent/src/provision.rs
```diff
@@ -321,49 +319,34 @@ pub async fn get_provision_status_wait(port: u16, duration: Option<Duration>) ->
...
-        return Err(Error::new(
-            ErrorKind::Other,
-            format!("Host response {} - {}", response.status, response_body),
-        ));
-    }
 
-    let state: ProivsionState = serde_json::from_str(&response_body)?;
-    Ok(state)
+    let mut headers = HashMap::new();
+    headers.insert(constants::METADATA_HEADER.to_string(), "true".to_string());
```
bingyuyin, line 338: does it miss a 'Host" header in the http request?

proxy_agent/src/proxy/proxy_authentication.rs
```diff
@@ -109,7 +109,7 @@ pub trait Authenticate {
     fn authenticate(
         &self,
         connection_id: u128,
-        request_url: String,
+        request_url: hyper::Uri,
```
bingyuyin, line 113: nit: update comments for the code example on top of this file

bingyuyin (dismissed):
approved with minor comments

proxy_agent/src/redirector.rs
```diff
@@ -55,8 +56,8 @@ pub struct AuditEntry {
     pub logon_id: u64,
     pub process_id: u32,
     pub is_admin: i32,
-    pub destination_ipv4: u32,
-    pub destination_port: u16,
+    pub destination_ipv4: u32, // in network byte order
```
LittleBoxOfSunshine, line 59: We're reliant on endianness here, I think it would be safer to have 4 u8s as a struct or array

proxy_agent/src/common/hyper_client.rs
```diff
@@ -475,6 +478,18 @@ pub fn full_body<T: Into<Bytes>>(chunk: T) -> BoxBody<Bytes, hyper::Error> {
         .boxed()
 }
 
+// check the request could skip the sig
+// and stream the body to the server directly
+pub fn should_skip_sig(method: hyper::Method, relative_uri: hyper::Uri) -> bool {
```
LittleBoxOfSunshine:
```suggestion
/// Certain endpoints are exempt from enforcement regardless of the VM's configuration.
/// Restricting access to these non-security impacting endpoints would introduce unreasonable
/// overhead and/or harm live-site investigations. Since the service won't require a signature,
/// there is no reason to generate one.
pub fn should_skip_sig(method: hyper::Method, relative_uri: hyper::Uri) -> bool {
```

This would make the context more clear imo

proxy_agent/src/common/hyper_client.rs
```diff
@@ -475,6 +478,18 @@ pub fn full_body<T: Into<Bytes>>(chunk: T) -> BoxBody<Bytes, hyper::Error> {
...
+    let url = relative_uri.to_string().to_lowercase();
+
+    // currently, we agreed to skip the sig for those requests:
+    //      o PUT   /vmAgentLog
+    //      o POST  /machine/?comp=telemetrydata
+    (method == hyper::Method::PUT || method == hyper::Method::POST)
+        && (url == "/machine/?comp=telemetrydata" || url == "/vmagentlog")
```
LittleBoxOfSunshine: This doesn't match your doc comment. It would be:
```suggestion
    (method == hyper::Method::PUT && url == "/vmagentlog")
    ||
    (method == hyper::Method::POST && url == "/machine/?comp=telemetrydata")
```

proxy_agent/src/key_keeper.rs
```diff
@@ -519,22 +519,23 @@ mod tests {
...
+        let shared_state = SharedState::new();
         // start wire_server listener
         let ip = "127.0.0.1";
         let port = 8081u16;
-
-        // LATER: need switch to tokio spawn once the server_mock is async
-        std::thread::spawn(move || {
-            server_mock::start(ip.to_string(), port);
-        });
+        let cloned_shared_state = shared_state.clone();
```
LittleBoxOfSunshine: Is this redundant? It gets cloned on final usage as well

proxy_agent/src/proxy/proxy_authentication.rs
```diff
@@ -313,9 +315,23 @@ pub fn get_authenticate(ip: String, port: u16, claims: Claims) -> Box<dyn Authen
...
+pub fn authenticate(
+    ip: String,
+    port: u16,
+    connection_id: u128,
+    request_uri: hyper::Uri,
+    claims: Claims,
+    shared_state: Arc<Mutex<SharedState>>,
+) -> bool {
+    let auth = get_authenticate(ip, port, claims);
+    Connection::write(connection_id, format!("Got auth: {}", auth.to_string()));
```
LittleBoxOfSunshine, line 328: Is there a better name for the enum than Authenticate? It would improve clarity, but also be easier to track safety. When I first read this line I was thinking we were sending a credential somewhere

proxy_agent/src/proxy/proxy_connection.rs
```diff
@@ -3,23 +3,40 @@
...
+    pub ip: Option<Ipv4Addr>, // currently, we only support IPv4
     pub port: u16,
 }
 
+impl ConnectionContext {
+    pub fn should_skip_sig(&self) -> bool {
+        hyper_client::should_skip_sig(self.method.clone(), self.url.clone())
+    }
+
+    pub fn request_ip(&self) -> String {
```
LittleBoxOfSunshine: Why string? `Option` would generally be better and should work if this is related to serialization or logging

LittleBoxOfSunshine (changes requested):

proxy_agent/src/provision.rs
ZhidongPeng: host header is common one, it will be added always in fn build_request.

proxy_agent/src/proxy/proxy_authentication.rs
ZhidongPeng: thanks, updated.

proxy_agent/src/redirector.rs
ZhidongPeng: I do not get what's the issue here, since it may not within the scope of this PR code changes, could you please help file an issue with more details? I will investigate further and may fix it later.

proxy_agent/src/common/hyper_client.rs
ZhidongPeng: Thanks, will update with your words.

proxy_agent/src/proxy/proxy_authentication.rs
ZhidongPeng: rename it to Authorizer and relative mod, trait and fn names.

proxy_agent/src/proxy/proxy_connection.rs
ZhidongPeng: this is for logging purposes. let me update the fn name and add bit more comments to it too.

proxy_agent/src/key_keeper.rs
ZhidongPeng: good catch.

proxy_agent/src/redirector.rs
ZhidongPeng: to add bit more context, this AuditEntry is for the struct ebpf map, it must match the order and size defined in the ebpf program.

bingyuyin (approved):

Issue #117, opened 2024-09-18
General idiom usage


TheSignPainter98:
Hi there! I took a quick look through this codebase and have a few friendly pointers to help get the most out of using Rust :) I think at the moment, much of the code I've seen feels as though it's been translated from another language such as C and although they can occupy much the same space, idiomatic Rust and C are quite different.

Please note that these things I spotted come from a sample of the files, so some impressions may be spurious. I also grouped these into one issue to avoid spam, so apologies for the length. Anyway, in no particular order, here's what I found:

- There seem to be quite a few places there's `_ = fallible_func()` is used, unexpectedly preventing error propagation. Whilst this may be okay in tests, it's definitely a code smell in general
- The proxy_agent cli argument parser is hand-rolled. Unless case insensitivity is required, I think it would be easier to maintain and more idiomatic to [derive a parser with `clap`](https://docs.rs/clap/latest/clap/_derive/_tutorial/chapter_0/index.html)
- Some fallible functions return booleans instead of a `Result`. This isn't idiomatic and has the added downside that the user can quietly ignore the error condition! By using a `Result` instead, the compiler will enforce that the error is considered, thus leading to more robust code. (e.g. [here](https://github.com/Azure/GuestProxyAgent/blob/e4de0187c817080099503e28a76134345d961b15/proxy_agent/src/redirector.rs#L95), I'd expect an error detailing what went wrong)
- I see very many free functions and although these are sometimes okay, most instances I've seen would be more idiomatically expressed as methods. E.g.
    ```rust
    fn attach_kprobe_program(bpf: &mut Bpf, ...)
    ```
    Would feel more idiomatic as
    ```rust
    impl Bpf {
        fn attach_kprobe_program(&mut self, ...)
    }
    ```
    It may be necessary to have a wrapper type for `Bpf` in this case as it is defined in the `aya` crate, but the newtype pattern, i.e. `struct Bpf(aya::Bpf);`, could help here
- Sometimes values where the type system (correctly) expresses invalid values get converted to valid-looking ones with invalid values. Specifically [here](https://github.com/Azure/GuestProxyAgent/blob/e4de0187c817080099503e28a76134345d961b15/proxy_agent_shared/src/misc_helpers.rs#L196), where if `proxy_agent_exe` is not valid utf8, `path_to_string` returns 'InvalidPath,' which the OS then attempts to execute.
    ```rust
    execute_command(&path_to_string(proxy_agent_exe), vec!["--version"], -1)
[...]

PR #119, opened 2024-09-18
handle ignored return values

ZhidongPeng:
This PR is addressing the first feedback from https://github.com/Azure/GuestProxyAgent/issues/117

`
There seem to be quite a few places there's _ = fallible_func() is used, unexpectedly preventing error propagation. Whilst this may be okay in tests, it's definitely a code smell in general
`


proxy_agent/src/main.rs
```diff
@@ -64,7 +64,7 @@ async fn main() {
             service::start_service(shared_state.clone());
             println!("Press Enter to end it.");
             let mut temp = String::new();
-            _ = std::io::stdin().read_line(&mut temp);
+            let _read = std::io::stdin().read_line(&mut temp);
```
shahneerali, line 67: does this get used?

shahneerali (approved):
approved with minor comment

proxy_agent/src/main.rs
ZhidongPeng: Not used on purpose, hence added under score.

PR #124, opened 2024-09-26
Use Azure SDK dev feed for v2 ProxyAgentSettings

ZhidongPeng:


LittleBoxOfSunshine (approved):

e2etest/GuestProxyAgentTest/Utilities/VMBuilder.cs
```diff
@@ -115,7 +115,15 @@ private async Task<VirtualMachineData> DoCreateVMData(ResourceGroupResource rgr,
                 {
                     ProxyAgentSettings = new ProxyAgentSettings()
                     {
-                        Enabled = true
+                        Enabled = true,
+                        Imds = new HostEndpointSettings()
+                        {
+                            InVmAccessControlProfileReferenceId= "/subscriptions/f21bdc74-62a2-437e-acbb-ecc81bbdee5c/resourceGroups/ProxyAgentE2E/providers/Microsoft.Compute/galleries/InVMAccessControlProfiles/inVMAccessControlProfiles/WindowsIMDS/versions/1.0.0",
```
bingyuyin: nit: {TestSetting.Instance.subscriptionId} as sub id, instead of hard code.

bingyuyin (dismissed):

e2etest/GuestProxyAgentTest/Utilities/VMBuilder.cs
ZhidongPeng: I thought about it, it is not just a sub-Id, it could be the whole uri since the profile uri could come from any sub and shared to our test sub.

bingyuyin (approved):

PR #130, opened 2024-10-11
Optimize the computed inVMAccessControlRules struct AuthorizationRules

ZhidongPeng:


proxy_agent/src/key_keeper/key.rs
```diff
@@ -40,6 +40,10 @@ const ENFORCE_MODE: &str = "enforce";
 //const ALLOW_DEFAULT_ACCESS: &str = "allow";
 //const DENY_DEFAULT_ACCESS: &str = "deny";
 
+pub trait Keyable {
+    fn get_key(&self) -> String;
```
mayankdaruka-msft: To implement a same trait on multiple structs, you can also define a custom macro to reduce duplicated code:

image

And then all you would need to write is:

`impl_Keyable!(Role, Privilege, Identity);`

proxy_agent/src/proxy/authorization_rules.rs
```diff
@@ -37,107 +29,198 @@ pub struct AuthorizationRules {
...
-                        match &access_control_rules.identities {
-                            Some(input_identities) => {
-                                for identity_name in role_assignment.identities {
-                                    for identity in input_identities {
-                                        if identity.name == identity_name {
-                                            identities.push(identity.clone());
-                                        }
-                                    }
+                                None => {
+                                    // skip the assignment if the role is not defined
```
LittleBoxOfSunshine: This should be unreachable, it's a sign that we were fed corrupted data. We should log errors in this branch

proxy_agent/src/proxy/authorization_rules.rs
```diff
@@ -37,107 +29,198 @@ pub struct AuthorizationRules {
...
+                        for assignment in role_assignments {
+                            match AuthorizationRules::get_from_dict(&roles, assignment.role.clone())
+                            {
+                                Some(role) => {
+                                    for privilege_name in role.privileges {
+                                        if AuthorizationRules::contains_key(
+                                            &privileges,
+                                            privilege_name.clone(),
+                                        ) {
+                                            let assignemnts = if privilege_assignments
```
LittleBoxOfSunshine: Typo `assignments`

proxy_agent/src/proxy/authorization_rules.rs
```diff
@@ -37,107 +29,198 @@ pub struct AuthorizationRules {
...
+                for v in vec {
+                    dict.insert(v.get_key(), v.clone());
+                }
+                Some(dict)
+            }
+            None => None,
+        }
+    }
+
+    fn contains_key<T>(dict: &Option<HashMap<String, T>>, key: String) -> bool {
```
LittleBoxOfSunshine: I don't think `contains_key` nor `get_from_dict` are necessary. These checks are only happening often if you aren't exiting this portion on the first case. Per the schema all of these values must be present in the json. If they aren't, it's corrupted data. We can log an error, and continue by acting as if no allow list was provided.

In general it's good, but especially in Rust it works well to validate data at the start and then represent your types as known good structs. Then there are no more options and thus no need for these helpers. If you are have to do too much error handling, it's usually a sign the data model should be changed.

proxy_agent/src/proxy/authorization_rules.rs
```diff
@@ -37,107 +29,198 @@ pub struct AuthorizationRules {
...
+    {
+        match dict {
+            Some(dict) => dict.get(&key).cloned(),
+            None => None,
         }
     }
 
     pub fn is_allowed(&self, connection_id: u128, request_url: hyper::Uri, claims: Claims) -> bool {
         if self.mode.to_lowercase() == "disabled" {
+            Connection::write_information(
```
LittleBoxOfSunshine, line 160: Storing `self.mode` as an enum will improve readability, get more safety checks, and a very minor performance improvement (cheaper comparison)

proxy_agent/src/proxy/authorization_rules.rs
```diff
@@ -37,107 +29,198 @@ pub struct AuthorizationRules {
...
-            let mut role_privilege_matched = false;
-            for rule in rules {
-                // is privilege match
-                for privilege in &rule.privileges {
-                    if privilege.is_match(connection_id, request_url.clone()) {
-                        role_privilege_matched = true;
-                        for identity in &rule.identities {
-                            if identity.is_match(connection_id, claims.clone()) {
-                                return true;
+        if let (Some(privileges), Some(identities), Some(privilege_assignments)) = (
```
LittleBoxOfSunshine: Storing these as a nested struct such as `AllowList` will enable you to use a single option, or generate a default list to use when parsing fails. Both would simplify the usage throughout.

proxy_agent/src/proxy/authorization_rules.rs
```diff
@@ -37,107 +29,198 @@ pub struct AuthorizationRules {
...
+            &self.privileges,
+            &self.identities,
+            &self.privilegeAssignments,
+        ) {
+            let mut privilege_matched = false;
+            for privilege in privileges.values() {
+                let privilege_name = privilege.name.clone();
```
LittleBoxOfSunshine: Is this clone needed?

proxy_agent/src/proxy/authorization_rules.rs
```diff
@@ -37,107 +29,198 @@ pub struct AuthorizationRules {
...
+                if privilege.is_match(connection_id, request_url.clone()) {
+                    privilege_matched = true;
+                    Connection::write_information(
+                        connection_id,
+                        format!("Reqeust matched privilege '{}'.", privilege_name),
+                    );
+
+                    if let Some(assignments) = privilege_assignments.get(&privilege_name) {
+                        for assignemnt in assignments {
```
LittleBoxOfSunshine: Typo assignment

proxy_agent/src/proxy/authorization_rules.rs
LittleBoxOfSunshine: This applies throughout. I suspect at least some of these clones are redundant and could be avoided with some minor parameter type tweaking.

proxy_agent/src/proxy/authorization_rules.rs
```diff
@@ -37,107 +29,198 @@ pub struct AuthorizationRules {
...
+            &self.privilegeAssignments,
+        ) {
+            let mut privilege_matched = false;
+            for privilege in privileges.values() {
+                let privilege_name = privilege.name.clone();
+                if privilege.is_match(connection_id, request_url.clone()) {
+                    privilege_matched = true;
+                    Connection::write_information(
+                        connection_id,
+                        format!("Reqeust matched privilege '{}'.", privilege_name),
```
LittleBoxOfSunshine: Spelling, Request

proxy_agent/src/proxy/authorization_rules.rs
```diff
@@ -37,107 +29,198 @@ pub struct AuthorizationRules {
...
+                    );
                 }
             }
 
-            if role_privilege_matched {
+            if privilege_matched {
                 Connection::write_information(
                     connection_id,
-                    "Privilege matched once, but no identity matches.".to_string(),
+                    "Privilege matched once, but no identity matches, deny the access.".to_string(),
```
LittleBoxOfSunshine: at least once

proxy_agent/src/proxy/authorization_rules.rs
```diff
@@ -37,107 +29,198 @@ pub struct AuthorizationRules {
...
+                } else {
+                    Connection::write_information(
+                        connection_id,
+                        format!("Reqeust does not match privilege '{}'.", privilege_name),
+                    );
                 }
             }
 
-            if role_privilege_matched {
+            if privilege_matched {
```
LittleBoxOfSunshine: Consider `any_privilege_matched` to communicate the intent more clearly

proxy_agent/src/proxy/authorization_rules.rs
```diff
@@ -37,107 +29,198 @@ pub struct AuthorizationRules {
...
                 );
                 return false;
             }
         }
 
         Connection::write_information(
             connection_id,
-            "No privilege matched, fall back to default access.".to_string(),
+            format!(
+                "No privilege matched, fall back to use the default access: {}.",
```
LittleBoxOfSunshine, line 230: I think this log is more verbose than it needs to be. Also, printing the boolean seems fine. If we want strings, it probably should be an enum with a ToString implementation

proxy_agent/src/proxy/authorization_rules.rs
```diff
@@ -37,107 +29,198 @@ pub struct AuthorizationRules {
...
+                                    Connection::write_information(
+                                        connection_id,
+                                        format!("Identity matched identity '{}'.", identity_name),
+                                    );
+                                    return true;
+                                } else {
+                                    Connection::write_information(
+                                        connection_id,
+                                        format!(
+                                            "Identity does not match identity '{}'.",
```
LittleBoxOfSunshine: It's too verbose to log each branch of each check, especially with full sentences. I think a summary at the end will be more readable for users.

It could cover the highlights like if default was used, which privilege matched if one matched. If none matched, we don't need each none printed it's implied by that high level summary

proxy_agent/src/key_keeper/key.rs
LittleBoxOfSunshine: I don't think we should create a trait for this. It doesn't have a clear conceptual purpose, we're just trying to reduce duplication presumably.

A better alternative would be to leverage the functional programming options. E.g.:

```rust
let privileges: HashMap<String, Privilege> = input_rules.privileges.iter()
        .map(|privilege| (privilege.name.to_string(), privilege))
        .collect();
```

proxy_agent/src/proxy/authorization_rules.rs
ZhidongPeng: renamed to `privilege_matched_once`

proxy_agent/src/proxy/authorization_rules.rs
ZhidongPeng: print the boolean directly.

proxy_agent/src/proxy/authorization_rules.rs
ZhidongPeng: removed.

proxy_agent/src/proxy/authorization_rules.rs
ZhidongPeng: make them all non-operational.

proxy_agent/src/proxy/authorization_rules.rs
ZhidongPeng: removed.

proxy_agent/src/proxy/authorization_rules.rs
ZhidongPeng: let's log it.

proxy_agent/src/proxy/authorization_rules.rs
ZhidongPeng: added an enum.

proxy_agent/src/proxy/authorization_rules.rs
ZhidongPeng: use borrowing as the rust error suggested

```
cannot move out of `privilege.name` which is behind a shared reference
move occurs because `privilege.name` has type `std::string::String`, which does not implement the `Copy` traitrustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20%5B1%5D?1#file%3A%2F%2F%2Fc%253A%2Fsrc%2Fgithub.com%2Fzhidongpeng%2FGuestProxyAgent%2Fproxy_agent%2Fsrc%2Fproxy%2Fauthorization_rules.rs)
authorization_rules.rs(170, 34): consider borrowing here: `&`
authorization_rules.rs(170, 48): consider cloning the value if the performance cost is acceptable: `.clone()`
```

LittleBoxOfSunshine (approved):

PR #133, opened 2024-10-15
Dump the AccessControl rules for supportability purpose

ZhidongPeng:
- Dump the AccessControl rules for supportability purpose in key_keeper module
- remove unnecessarily ownership move in misc_helper function json_write_to_file & json_read_from_file

### DIFF
--- a/proxy_agent/src/key_keeper.rs
+++ b/proxy_agent/src/key_keeper.rs
@@ -27,8 +27,11 @@ pub mod key;
 use self::key::Key;
 use crate::common::{constants, helpers, logger};
 use crate::provision;
+use crate::proxy::authorization_rules::{AuthorizationRulesForLogging, ComputedAuthorizationRules};
 use crate::proxy::proxy_authorizer;
-use crate::shared_state::{key_keeper_wrapper, tokio_wrapper, SharedState};
+use crate::shared_state::{
+    key_keeper_wrapper, proxy_authenticator_wrapper, tokio_wrapper, SharedState,
+};
 use crate::{acl, redirector};
 use hyper::Uri;
 use proxy_agent_shared::misc_helpers;
@@ -52,12 +55,14 @@ const DELAY_START_EVENT_THREADS_IN_MILLISECONDS: u128 = 60000; // 1 minute
 /// poll secure channel status at interval
 ///  - base_url: the WireServer endpoint base url
 ///  - key_dir: the folder to save the key details
+///  - log_dir: the folder to log the access control rule details
 ///  - interval: the interval to poll the secure channel status
 ///  - config_start_redirector: boolean to indicate start the redirector when the key keeper task is running
 ///  - shared_state: the global shared state
 pub async fn poll_secure_channel_status(
     base_url: Uri,
     key_dir: PathBuf,
+    log_dir: PathBuf,
     interval: Duration,
     config_start_redirector: bool,
     shared_state: Arc<Mutex<SharedState>>,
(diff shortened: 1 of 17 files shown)


proxy_agent_shared/src/misc_helpers.rs
```diff
@@ -61,7 +62,7 @@ pub fn try_create_folder(dir: PathBuf) -> std::io::Result<()> {
     Ok(())
 }
 
-pub fn json_write_to_file<T>(obj: &T, file_path: PathBuf) -> std::io::Result<()>
+pub fn json_write_to_file<T>(obj: &T, file_path: &PathBuf) -> std::io::Result<()>
 where
```
mayankdaruka-msft, line 70: It would generally be better practice to pass in `&Path` to functions that don't need to modify the path, instead of `&PathBuf`.

mayankdaruka-msft: `&Path` will accept arguments of both types: `&Path` and `&PathBuf`, so no other changes in codebase is needed. However it is better to make the change here to enforce that we are not modifying the path.

proxy_agent/src/key_keeper.rs
```diff
@@ -238,6 +247,20 @@ async fn loop_poll(
...
+            let rules = AuthorizationRulesForLogging::new(
+                status.authorizationRules.clone(),
+                ComputedAuthorizationRules {
+                    wireserver: proxy_authenticator_wrapper::get_wireserver_rules(
+                        shared_state.clone(),
+                    ),
+                    imds: proxy_authenticator_wrapper::get_imds_rules(shared_state.clone()),
+                },
+            );
+            rules.write_all(&log_dir, 20);
```
mayankdaruka-msft: Define a constant for 20?

proxy_agent_shared/src/misc_helpers.rs
```diff
@@ -165,6 +166,53 @@ pub fn get_files(dir: &PathBuf) -> std::io::Result<Vec<PathBuf>> {
...
+/// use std::path::PathBuf;
+/// use proxy_agent_shared::misc_helpers;
+/// let dir = PathBuf::from("C:\\");
+/// let search_regex_pattern = r"^(.*\.log)$";  // search for files with .log extension
+/// let files = misc_helpers::search_files(&dir, search_regex_pattern).unwrap();
+///
+/// let search_regex_pattern = r"^MyFile.*\.json$"; // Regex pattern to match "MyFile*.json"
+/// let files = misc_helpers::search_files(&dir, search_regex_pattern).unwrap();
+/// ```
+pub fn search_files(dir: &PathBuf, search_regex_pattern: &str) -> std::io::Result<Vec<PathBuf>> {
```
mayankdaruka-msft: It's also worth considering switching over to use `&Path` here instead of `PathBuf` since it looks like we don't need to modify the paths we are retrieving.

proxy_agent_shared/src/misc_helpers.rs
```diff
@@ -165,6 +166,53 @@ pub fn get_files(dir: &PathBuf) -> std::io::Result<Vec<PathBuf>> {
...
+    };
+
+    for entry in fs::read_dir(dir)? {
+        let entry = entry?;
+        let file_full_path = entry.path();
+        let metadata = fs::metadata(&file_full_path)?;
+        if !metadata.is_file() {
+            continue;
+        }
+        let file_name = get_file_name(file_full_path.clone());
```
mayankdaruka-msft, line 211: Can we pass reference instead of cloning?

ZhidongPeng: This change will come in next PR.

proxy_agent_shared/src/misc_helpers.rs
ZhidongPeng: yeah.

mayankdaruka-msft (approved):

PR #134, opened 2024-10-15
Proxy Agent: Introduce custom result and error types

mayankdaruka-msft:
This PR is addressing the feedback from #117.
```
I see std::io::Result being used as a custom result type (a common rookie error seen for example [here](https://github.com/Azure/GuestProxyAgent/blob/e4de0187c817080099503e28a76134345d961b15/proxy_agent/src/common/hyper_client.rs#L256-L257)), but really crates should define their own error and result types, leading to a nice 'branded' feeling API (e.g. in crate foo, foo::bar() might return a foo::Result which can hold a foo::Error). This is surprisingly straightforward to remedy, the usual way this is done is by using thiserror deriving a custom error type and making a new Result alias as follows (I like to put these next to one another in lib.rs
#[derive(Debug, thiserror::Error)]
pub enum Error {
    #[error("cannot foo")]
    Foo,
}

pub type Result<T> = std::result::Result<T, Error>
Doing this also has the added benefit of encouraging better error handling
```

Throughout the codebase, we see the usage of `Result` and `Error` types from the `std::io` library. Most of those errors are not I/O related and needed some re-defining. In this PR, we introduce our own type to be used in the proxy_agent crate.

The changes here do not move over all occurrences of std::io::Result usage in proxy agent code. There will be a subsequent PR that builds on top of the work here so we can keep the PRs smaller.

E2E Tests Link: https://msazure.visualstudio.com/One/_build/results?buildId=105933413&view=results


proxy_agent/src/common/error.rs
```diff
@@ -0,0 +1,152 @@
+// Copyright (c) Microsoft Corporation
+// SPDX-License-Identifier: MIT
+
+use std::error::Error as StdError;
+use std::fmt::Display;
+
```
ZhidongPeng, line 7: nit: remove extra line break. it would be nice keep all the use statements together.

proxy_agent/Cargo.toml
```diff
@@ -26,6 +26,7 @@ hyper-util = { version = "0.1", features = ["tokio"] }
 tower = { version = "0.4", features = ["full"] }
 tower-http = { version = "0.5.2", features = ["limit"] }
 clap = { version = "4.5.17", features =["derive"] } # Command Line Argument Parser 
+thiserror = "1.0.64"
```
ZhidongPeng, line 29: Cargo.lock file missed?

ZhidongPeng (changes requested):

proxy_agent/Cargo.toml
LittleBoxOfSunshine: That reminds me, if we add [--locked](https://doc.rust-lang.org/cargo/commands/cargo-build.html#manifest-options) to the cargo build step in the pipeline that will catch this error and increases security + determinism.

proxy_agent/src/common/error.rs
mayankdaruka-msft: Done, thanks.

ZhidongPeng (approved):

PR #136, opened 2024-10-17
Add Check Spelling action and fixing typos

ZhidongPeng:


.github/actions/spelling/expect.txt
```diff
@@ -0,0 +1,329 @@
...
+ccbf
+cicd
+cimv
+cla
+cmds
+codeofconduct
+codeql
+collectguestlogs
+commandline
+commom
```
mayankdaruka-msft: This is not an expected word right? Are there similar ones in this file that are not expected?

ZhidongPeng: these are expected words actually, it is more like our own dictionary.

ZhidongPeng: `commom` is a typo, I need remove it from expect.txt and fix this typo.

shahneerali (approved):

PR #138, opened 2024-10-18
Proxy Agent: Convert all remaining errors to custom-defined error types

mayankdaruka-msft:
This PR is addressing the feedback from https://github.com/Azure/GuestProxyAgent/issues/117.

This PR is a follow up to my previous PR - https://github.com/Azure/GuestProxyAgent/pull/134 - where I introduced custom result and error types as part of improving overall readability, maintainability and error handling within our codebase.

In this PR, all remaining error instances are moved over to custom error types defined within the proxy_agent crate only. Since this PR is already large enough, I will introduce changes to the extension and shared crates in a next PR.

Tests Link: https://msazure.visualstudio.com/One/_build/results?buildId=106290309&view=results (150/150 passed)

### DIFF
--- a/proxy_agent/src/common/error.rs
+++ b/proxy_agent/src/common/error.rs
@@ -40,6 +40,14 @@ impl Error {
     pub fn wire_server(error_type: WireServerErrorType, message: String) -> Self {
         Self::new(ErrorType::WireServer(error_type, message))
     }
+
+    pub fn acl(error: AclErrorType, error_code: u32) -> Self {
+        Self::new(ErrorType::Acl(error, error_code))
+    }
+
+    pub fn bpf(error: BpfErrorType) -> Self {
+        Self::new(ErrorType::Bpf(error))
+    }
 }
 
 impl Display for Error {
@@ -69,6 +77,12 @@ enum ErrorType {
 
     #[error("Failed to parse URL {0} with error: {1}")]
     ParseUrl(String, String),
+
+    #[error("acl_directory: {0} with error: {1}")]
+    Acl(AclErrorType, u32),
+
+    #[error("{0}")]
+    Bpf(BpfErrorType),
 }
 
 #[derive(Debug, thiserror::Error)]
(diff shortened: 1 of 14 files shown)


proxy_agent/src/common/error.rs
```diff
@@ -113,6 +141,69 @@ pub enum KeyErrorType {
...
+    #[error("Loading NetUserGetLocalGroups failed with error: {0}")]
+    LoadNetUserGetLocalGroups(String),
+
+    #[error("LsaGetLogonSessionData {0}")]
+    LsaGetLogonSessionData(String),
+
+    #[error("WinSock::WSAIoctl - SIO_QUERY_WFP_CONNECTION_REDIRECT_CONTEXT: {0}")]
+    WSAIoctl(i32),
+
+    #[error("GlobalMemoryStatusExfailed: {0}")]
```
ZhidongPeng: Try rename it `GlobalMemoryStatusEx failed`

proxy_agent/src/proxy/proxy_server.rs
```diff
@@ -257,16 +270,10 @@ async fn handle_request(
...
-                            let err = format!("Failed to get lookup_audit_from_stream: {}", e);
-                            event_logger::write_event(
-                                event_logger::WARN_LEVEL,
-                                err,
-                                "handle_request",
-                                "proxy_listener",
-                                Connection::CONNECTION_LOGGER_KEY,
-                            );
-                        }
+                        Connection::write_warning(
```
ZhidongPeng: we need `event_logger::write_event` since we want this error send to kusto.

ZhidongPeng (changes requested):

proxy_agent/src/proxy/proxy_server.rs
mayankdaruka-msft: There is no code path resulting in the "Unsupported" error kind. This was the case even before my changes, AFAIK, which is why I removed it.

There was a function `get_audit_from_stream` which returned unsupported on Linux, but that's not being called and I've removed that function in this PR as well. I'm assuming this logic to send the Kusto event was added when that function was being called, but doesn't look relevant anymore.

mayankdaruka-msft: My mistake, just pushed a fix.

ZhidongPeng (approved):

PR #142, opened 2024-10-23
Move free functions for BpfObject method

ZhidongPeng:
This PR is addressing the feedback from https://github.com/Azure/GuestProxyAgent/issues/117

* add instance functions for Windows BpfObject.
* add instance functions for Linux BpfObject

### DIFF
--- a/proxy_agent/src/redirector/linux.rs
+++ b/proxy_agent/src/redirector/linux.rs
@@ -23,19 +23,22 @@ use std::convert::TryFrom;
 use std::path::PathBuf;
 use std::sync::{Arc, Mutex};
 
-pub type BpfObject = Bpf;
+pub struct BpfObject(Bpf);
 
 pub fn start_internal(local_port: u16, shared_state: Arc<Mutex<SharedState>>) -> bool {
-    let mut bpf = match open_ebpf_file(super::get_ebpf_file_path(), shared_state.clone()) {
-        Some(value) => value,
-        None => return false,
+    let mut bpf_object = match BpfObject::from_ebpf_file(super::get_ebpf_file_path()) {
+        Ok(value) => value,
+        Err(e) => {
+            set_error_status(format!("{}", e), shared_state.clone());
+            return false;
+        }
     };
 
-    for (name, _map) in bpf.maps() {
+    for (name, _map) in bpf_object.0.maps() {
         logger::write(format!("found map '{}'", name));
     }
 
-    for (name, prog) in bpf.programs() {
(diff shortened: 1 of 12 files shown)


shahneerali (dismissed):

proxy_agent/src/redirector/linux.rs
```diff
@@ -285,236 +217,166 @@ fn update_policy_map(
...
                         }
                     }
                 }
                 Err(err) => {
-                    set_error_status(
-                        format!("Failed to load HashMap 'policy_map' with error: {}", err),
-                        shared_state.clone(),
-                    );
-                    return false;
+                    return Err(Error::general(format!(
```
mayankdaruka-msft: Are all of these errors general? I think we should create more focused and specific error types because this helps with readability. We should also reuse existing error types if applicable.

For example. for this error we can use the LoadBpfMapHashMap error, but alter it so it takes in the name of the map we failed to load.

proxy_agent/src/redirector/linux.rs
```diff
@@ -102,138 +109,68 @@ pub fn start_internal(local_port: u16, shared_state: Arc<Mutex<SharedState>>) ->
...
-
-        if let Some(addr) = nic.address {
-            if let Some(socket_addr) = addr.as_sockaddr_in() {
-                return Some(socket_addr.ip().to_string());
+                    )));
+                }
+            },
+            None => {
+                return Err(Error::general(
+                    "Failed to get map 'skip_process_map'.".to_string(),
```
mayankdaruka-msft: Similarly can use GetBpfMap here but would have to introduce a new property for map name.

proxy_agent/src/redirector/linux.rs
```diff
@@ -285,236 +217,166 @@ fn update_policy_map(
...
                     match program.load() {
-                        Ok(_) => logger::write("connect4 program loaded.".to_string()),
+                        Ok(_) => logger::write("tcp_v4_connect program loaded.".to_string()),
                         Err(err) => {
-                            let message =
-                                format!("Failed to load program 'connect4' with error: {}", err);
-                            set_error_status(message.to_string(), shared_state.clone());
-                            return false;
+                            return Err(Error::general(format!(
+                                "Failed to load program 'tcp_v4_connect' with error: {}",
```
mayankdaruka-msft: I see a lot of similar errors: "Failed to load program....", "Failed to get program...", "Failed to attach program..."
They can all come under a more specific type of error.

mayankdaruka-msft (approved):

PR #143, opened 2024-10-25
Proxy Agent: Move custom error type to enum

mayankdaruka-msft:
This PR is addressing feedback from #117.

In my previous PRs, I introduced the custom error as a struct with a boxed error type enum. However since we have a finite set of error types, it would be more efficient to handle this during compile time via an enum directly instead of requiring dynamically sized error handling with a box.

This PR moves to using an enum for the error type defined in the proxy agent crate. Other crates will use a similar pattern.

### DIFF
--- a/proxy_agent/src/host_clients/wire_server_client.rs
+++ b/proxy_agent/src/host_clients/wire_server_client.rs
@@ -27,7 +27,7 @@ use crate::common::{
 };
 use crate::host_clients::goal_state::{GoalState, SharedConfig};
 use crate::shared_state::{key_keeper_wrapper, SharedState};
-use http::Method;
+use http::{uri::InvalidUri, Method};
 use hyper::Uri;
 use std::collections::HashMap;
 use std::sync::{Arc, Mutex};
@@ -56,7 +56,9 @@ impl WireServerClient {
         }
 
         let url = format!("http://{}:{}/{}", self.ip, self.port, TELEMETRY_DATA_URI);
-        let url: Uri = url.parse().map_err(|e| Error::parse_url(url, e))?;
+        let url: Uri = url
+            .parse()
+            .map_err(|e: InvalidUri| Error::ParseUrl(url, e.to_string()))?;
         let mut headers = HashMap::new();
         headers.insert("x-ms-version".to_string(), "2012-11-30".to_string());
         headers.insert(
@@ -78,7 +80,7 @@ impl WireServerClient {
             {
                 Ok(r) => r,
                 Err(e) => {
-                    return Err(Error::wire_server(
+                    return Err(Error::WireServer(
                         WireServerErrorType::Telemetry,
                         format!("Failed to send request {}", e),
                     ))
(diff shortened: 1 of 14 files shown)


proxy_agent/src/proxy/proxy_server.rs
```diff
@@ -193,7 +193,7 @@ async fn accept_one_request(
 fn set_stream_read_time_out(stream: TcpStream) -> Result<(TcpStream, std::net::TcpStream)> {
     // Convert the stream to a std stream
     let std_stream = stream.into_std().map_err(|e| {
-        Error::io(
+        Error::IO(
```
LittleBoxOfSunshine: In general, I recommend treating acronyms as a single token like a word. It avoids debates and confusion over what's a word and what isn't like TV. Idk what the project has standardized on if anything. I know in .NET there is an exception for 2 letter ones, but A) that's unreasonable and B) this isn't .NET so I would encourage adoption `Io` here.

LittleBoxOfSunshine: Also, looks like it's not just me it's in the RFC for rust https://rust-lang.github.io/api-guidelines/naming.html#:~:text=Basic%20Rust%20naming%20conventions%20are%20described%20in%20RFC%20430.&text=snake_case!&text=In%20UpperCamelCase%20%2C%20acronyms%20and%20contractions,are%20lower%2Dcased%3A%20is_xid_start%20.

ZhidongPeng: I think he means this sentence from the link above:
`In UpperCamelCase, acronyms and contractions of compound words count as one word: use Uuid rather than UUID, Usize rather than USize or Stdin rather than StdIn.`

mayankdaruka-msft: Thanks, resolved.

proxy_agent/src/host_clients/wire_server_client.rs
```diff
@@ -56,7 +56,9 @@ impl WireServerClient {
         }
 
         let url = format!("http://{}:{}/{}", self.ip, self.port, TELEMETRY_DATA_URI);
-        let url: Uri = url.parse().map_err(|e| Error::parse_url(url, e))?;
+        let url: Uri = url
+            .parse()
+            .map_err(|e: InvalidUri| Error::ParseUrl(url, e.to_string()))?;
```
ZhidongPeng: I see you explicitly add the error type for InvalidUri in many places. is it on purpose and required to make code built?

ZhidongPeng: Please update them like 
`        let url = url
            .parse::<Uri>()
            .map_err(|e| Error::ParseUrl(url, e.to_string()))?;`

ZhidongPeng (changes requested):

proxy_agent/src/host_clients/wire_server_client.rs
mayankdaruka-msft: Done, thanks.

ZhidongPeng (approved):

PR #144, opened 2024-10-25
start proxy listener with retry logic if the address is in use

ZhidongPeng:


bingyuyin (dismissed):

proxy_agent/src/proxy/proxy_server.rs
```diff
@@ -68,6 +69,38 @@ pub fn get_status(shared_state: Arc<Mutex<SharedState>>) -> ProxyAgentDetailStat
     }
 }
 
+/// start listener at the given address with retry logic if the address is in use
+async fn start_listener_with_retry(addr: &str, retry_count: u16) -> std::io::Result<TcpListener> {
```
mayankdaruka-msft: This function should return the custom Result type.

ZhidongPeng: no need custom error type at this place, it returns it is own bind error, which is std::io::Result<TcpListener>

ZhidongPeng: replaced with our own Io error

proxy_agent/src/proxy/proxy_server.rs
```diff
@@ -68,29 +69,71 @@ pub fn get_status(shared_state: Arc<Mutex<SharedState>>) -> ProxyAgentDetailStat
...
+}
+
+pub async fn start(port: u16, shared_state: Arc<Mutex<SharedState>>) {
     Connection::init_logger(config::get_logs_dir());
 
     let addr = format!("{}:{}", std::net::Ipv4Addr::LOCALHOST, port);
     logger::write(format!("Start proxy listener at '{}'.", &addr));
 
-    let listener = match TcpListener::bind(&addr).await {
+    let listener = match start_listener_with_retry(&addr, 5).await {
```
mayankdaruka-msft, line 115: Can you make this a constant? https://doc.rust-lang.org/rust-by-example/custom_types/constants.html

proxy_agent/src/proxy/proxy_server.rs
```diff
@@ -68,29 +69,71 @@ pub fn get_status(shared_state: Arc<Mutex<SharedState>>) -> ProxyAgentDetailStat
     }
 }
 
-pub async fn start(
-    port: u16,
-    shared_state: Arc<Mutex<SharedState>>,
-) -> core::result::Result<(), Box<dyn std::error::Error + Send + Sync>> {
+/// start listener at the given address with retry logic if the address is in use
+async fn start_listener_with_retry(addr: &str, retry_count: u16) -> Result<TcpListener> {
+    let sleep_duration = Duration::from_secs(1);
```
mayankdaruka-msft, line 74: Same for this.

proxy_agent/src/proxy/proxy_server.rs
```diff
@@ -68,29 +69,71 @@ pub fn get_status(shared_state: Arc<Mutex<SharedState>>) -> ProxyAgentDetailStat
...
 
     let addr = format!("{}:{}", std::net::Ipv4Addr::LOCALHOST, port);
     logger::write(format!("Start proxy listener at '{}'.", &addr));
 
-    let listener = match TcpListener::bind(&addr).await {
+    let listener = match start_listener_with_retry(&addr, 5).await {
         Ok(listener) => listener,
         Err(e) => {
-            let message = format!("Failed to bind TcpListener '{}' with error {}.", addr, e);
+            let message = format!("{}", e);
```
mayankdaruka-msft, line 118: Nit: For single value conversions using `to_string()` is the more idiomatic and efficient approach. `format!` is better for more complex formatting.

mayankdaruka-msft (approved):

PR #146, opened 2024-10-28
Pass environment for the linux_ebpf_test

ZhidongPeng:
linux_ebpf_test requires root permission and BPF capability to run, so far, we know some container build environments do not have BPF capability, this test will skip if the current environment does not have the capability to load BPF programs.
It used to be determined by the /.dockerenv file exists, it apparently not always true.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -176,6 +176,7 @@ norestart
 ntdll
 NTSTATUS
 onscreen
+onebranch
 opencode
 opensource
 percpu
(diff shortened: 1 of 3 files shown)


mayankdaruka-msft (commented):
Is there another approach we can use to determine if an image  has BPF capabilities?
Maybe within the test we can attempt running a basic BPF operation to see if it works, and if it doesn't then the image doesn't have the capability, and we don't run the remaining part of the test?

shahneerali (approved):

PR #155, opened 2024-11-19
Test: if the reboot operation failed, try check the VM instance view

ZhidongPeng:


shahneerali (approved):
I believe E2E token has 2-hour timeout. Upper bound of E2E runs is 1 hours 20 minutes so will have to keep in mind the total time for E2E tests if we introduce more retries for other functions in the future.

PR #183, opened 2025-01-24
Fixes for less permissive identity matching

mayankdaruka-msft:
- Removes case insensitive checks
- Stores process name and path in OsString and PathBuf (OsString underlying) respectively, which are designed to hold data that does not necessarily conform to UTF-8. This prevents the need for lossy conversion which could potentially result in incorrect identity matching

### DIFF
--- a/proxy_agent/src/key_keeper/key.rs
+++ b/proxy_agent/src/key_keeper/key.rs
@@ -300,7 +300,7 @@ impl Identity {
             format!("Start to match identity '{}'", self.name),
         );
         if let Some(ref user_name) = self.userName {
-            if user_name.to_lowercase() == claims.userName.to_lowercase() {
+            if user_name.to_string() == claims.userName {
                 logger.write(
                     LoggerLevel::Verbose,
                     format!(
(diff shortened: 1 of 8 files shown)


proxy_agent/src/proxy/proxy_summary.rs
```diff
@@ -50,7 +52,7 @@ impl From<ProxySummary> for ProxyConnectionSummary {
             userGroups: Some(proxy_summary.userGroups.clone()),
             ip: proxy_summary.ip.to_string(),
             port: proxy_summary.port,
-            processFullPath: Some(proxy_summary.processFullPath.to_string()),
+            processFullPath: Some(proxy_summary.processFullPath.to_string_lossy().to_string()),
```
ZhidongPeng, line 55: do we need to_string() after to_string_lossy()?

proxy_agent/src/key_keeper/key.rs
```diff
@@ -1313,6 +1319,15 @@ mod tests {
...
         );
+        let identity3 = r#"{
+            "name": "test",
+            "processName": "Test"
+        }"#;
+        let identity3: Identity = serde_json::from_str(identity3).unwrap();
+        assert!(
+            !identity3.is_match(&logger, &claims),
+            "identity should not be matched"
+        );
```
ZhidongPeng, line 1335: Please add UT coverage for non-ut8 characters too.

proxy_agent/src/proxy/proxy_summary.rs
mayankdaruka-msft: Yeah, to_string_lossy gives us a `Cow<str>`. We need to convert into `String`.

ZhidongPeng (approved):

PR #188, opened 2025-02-04
enable systemd sandboxing features

nmeyerhans:
This change adjusts service relationships to more common targets and dependencies and enables a number of systemd sandboxing features.  As the GPA otherwise runs as root without any confiment, this is adds additional layers of defense in depth.

Tested on Debian unstable with systemd 257.2 on both arm64 and amd64. With the change in place, `systemd-analyze security azure-guest-proxy-agent.service` reports
> `Overall exposure level for azure-guest-proxy-agent.service: 5.2 MEDIUM 😐`

versus 

> `Overall exposure level for azure-guest-proxy-agent.service: 9.6 UNSAFE 😨`

without sandboxing.


proxy_agent_setup/src/linux/azure-proxy-agent.service
```diff
@@ -1,6 +1,5 @@
 [Unit]
 Description=Microsoft Azure GuestProxyAgent
-DefaultDependencies=no
```
cjp256, line 3: This breaks on Ubuntu 22.04.  Not everywhere but at least that.

[    3.097257] temp-vm-cpatterson-eastus2-t02061450001 systemd[1]: sysinit.target: Found ordering cycle on cloud-init.service/start
[    3.099890] temp-vm-cpatterson-eastus2-t02061450001 systemd[1]: sysinit.target: Found dependency on systemd-networkd-wait-online.service/start
[    3.102431] temp-vm-cpatterson-eastus2-t02061450001 systemd[1]: sysinit.target: Found dependency on systemd-networkd.service/start
[    3.104923] temp-vm-cpatterson-eastus2-t02061450001 systemd[1]: sysinit.target: Found dependency on network-pre.target/start
[    3.107574] temp-vm-cpatterson-eastus2-t02061450001 systemd[1]: sysinit.target: Found dependency on azure-proxy-agent.service/start
[    3.110364] temp-vm-cpatterson-eastus2-t02061450001 systemd[1]: sysinit.target: Found dependency on sysinit.target/start
[    3.112499] temp-vm-cpatterson-eastus2-t02061450001 systemd[1]: sysinit.target: Job cloud-init.service/start deleted to break ordering cycle starting with sysinit.target/start

proxy_agent_setup/src/linux/azure-proxy-agent.service
```diff
@@ -11,5 +10,45 @@ ExecStart=/usr/sbin/azure-proxy-agent
 Restart=always
 RestartSec=5
 
+ProtectSystem=strict
```
cjp256, line 13: Sandboxing looks great, but I would make sure that the extension variant is tested along with the self-update process and make sure the sandbox requirements don't explode.

Otherwise I expect sytemd will ignore (though complain about) any flags a particular distro's systemd doesn't have support for.

proxy_agent_setup/src/linux/azure-proxy-agent.service
cjp256: Is there a downside to keeping this flag in place (even if not ideal?)

nmeyerhans: > Is there a downside to keeping this flag in place (even if not ideal?)

I'd flip the question around, and ask exactly what it is in DefaultDependencies you need to avoid? Per the documentation, the default dependencies "ensure that normal service units pull in basic system initialization, and are terminated cleanly prior to system shutdown.. Only services involved with early boot or late system shutdown should disable this option."  That does not describe the GPA. So if the default dependencies are causing problems, it's something worth looking into. It likely indicates a problem elsewhere.

proxy_agent_setup/src/linux/azure-proxy-agent.service
ZhidongPeng: We will run extension private build E2E tests.

ZhidongPeng (approved):

PR #201, opened 2025-02-26
Logging Infra refinement

ZhidongPeng:
- Introduce `log` crate for logging level
- Introduce `ctor` crate to setup and clean up loggers once, no need to setup the logger for every unit tests
- Removed the mpsc/async mode file logging.
- Instead of Acl all ProxyAgent folders in windows, acl keys and executable folders
- Change file log level in config to `Trace` to log all logs for E2E tests.

### DIFF
--- a/proxy_agent/src/service.rs
+++ b/proxy_agent/src/service.rs
@@ -63,7 +63,7 @@ pub async fn start_service(shared_state: SharedState) {
     });
 }
 
-pub fn setup_loggers(log_folder: PathBuf, max_logger_level: LoggerLevel) {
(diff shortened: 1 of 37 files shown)


proxy_agent/src/service.rs
```diff
@@ -68,6 +63,30 @@ pub async fn start_service(shared_state: SharedState) {
     });
 }
 
+pub fn setup_loggers(log_folder: PathBuf, max_logger_level: LoggerLevel) {
```
ZhidongPeng: no need to be pub

shahneerali (approved):

PR #208, opened 2025-03-13
escape xml characters when write status.tag file

ZhidongPeng:


proxy_agent/src/provision.rs
```diff
@@ -425,6 +425,11 @@ async fn write_provision_state(
         }
     }
 
+    if !failed_state_message.is_empty() {
+        // escape xml characters to allow the message to able be composed into xml payload
+        failed_state_message = helpers::xml_escape(failed_state_message);
```
privateHmmmm, line 430: Can you please add some unit test for it please, to make sure future code won't break this contract

huiminyan2017, line 431: Can you please add some unit test for it please, to ensure future code won't break this contract

ZhidongPeng: added

huiminyan2017 (approved):

bingyuyin (approved):

PR #248, opened 2025-06-25
GPA writes windows events to 'Windows Azure' instead of 'Application'.

ZhidongPeng:
1. Rename application.rs file to etw_writter.rs
2. Rename ApplicationEventWritter to WindowsEventWritter and accept the event log name instead of hard coded to `Application'
3. Extract test mod etw_reader to etw_reader.rs file
4. Windows GPA service passes `Windows Azure` as event log name during init.


bingyuyin (approved):

proxy_agent_shared/src/logger/logger_manager.rs
```diff
@@ -65,23 +65,23 @@ pub fn set_loggers(
...
+                    if let Err(e) = WINDOWS_ETW_LOGGER.set(logger) {
                         write_system_log(
                             LoggerLevel::Error,
-                            format!("Failed to set Windows Application ETW logger: {e}"),
+                            format!("Failed to set Windows ETW logger: {e}"),
                         );
                     }
                 }
                 Err(e) => {
                     write_system_log(
```
bingyuyin, line 82: in case of etw setup failed/error, do you need to log local file instead?

ZhidongPeng: kind of good idea, let's see if we need it or not in future.

PR #260, opened 2025-07-24
move update_current_secure_channel_state before reporting attest the key

ZhidongPeng:
It is a race condition bug, when key_latched and all provisioned, GPA starts the status reporting async task (_some other telemetry async tasks too_), and then update the secure channel state into shared state. There is a risk, status reporting task could get the current_secure_channel_state from shared state earlier than GPA updated it. 
Hence there will be 1 minutes delay our status reporting task to get the updated current_secure_channel_state.

### DIFF
--- a/proxy_agent/src/key_keeper.rs
+++ b/proxy_agent/src/key_keeper.rs
@@ -397,6 +397,11 @@ impl KeyKeeper {
             }
 
             let state = status.get_secure_channel_state();
+            let secure_channel_state_updated = self
+                .key_keeper_shared_state
+                .update_current_secure_channel_state(state.to_string())
+                .await;
+
             // check if need fetch the key
             if state != DISABLE_STATE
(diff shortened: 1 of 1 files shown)


proxy_agent/src/key_keeper.rs
```diff
@@ -397,6 +397,11 @@ impl KeyKeeper {
             }
 
             let state = status.get_secure_channel_state();
```
huiminyan2017, line 399: This loop_poll function is too long for reviewers to read and maintenance. Consider refactor it when you can

huiminyan2017 (approved):

proxy_agent/src/key_keeper.rs
ZhidongPeng: yeah, will try break down to sub function later.

PR #272, opened 2025-09-05
Add `DefaultDependencies=no` back for Linux GPA service

ZhidongPeng:


gjolly:
I don't think you can comment like this in a unit file. Here is the error I got when trying to test this change:

```
[    8.647653] systemd[1]: /usr/lib/systemd/system/azure-proxy-agent.service:4: Failed to parse boolean value, ignoring: no      # azure-proxy-agent supposed to be an early boot and late system shutdown service
```

gjolly:
Otherwise, I tested the change without the comment and it seems to fix the issue we were seeing on Ubuntu 24.04 LTS.

ZhidongPeng:
> I don't think you can comment like this in a unit file. Here is the error I got when trying to test this change:
> 
> ```
> [    8.647653] systemd[1]: /usr/lib/systemd/system/azure-proxy-agent.service:4: Failed to parse boolean value, ignoring: no      # azure-proxy-agent supposed to be an early boot and late system shutdown service
> ```

Thanks for it, I fixed and tested it now.

proxy_agent_setup/src/linux/azure-proxy-agent.service
```diff
@@ -1,5 +1,7 @@
 [Unit]
 Description=Microsoft Azure GuestProxyAgent
+# azure-proxy-agent supposed to be an early boot and late system shutdown service
+DefaultDependencies=no
```
nmeyerhans, line 4: If you add this, I think you'll also want `Conflicts=shutdown.target` and `Before=shutdown.target` to ensure the service is properly stopped on shutdown or reboot.

And I think you will want `After=sysinit.target` to ensure that basic system initialization has completed, but it's worth a careful read of [systemd.special(7)](https://manpages.debian.org/trixie/systemd/systemd.special.7.en.html) and [bootup(7)](https://manpages.debian.org/trixie/systemd/bootup.7.en.html) to be sure.

ZhidongPeng:
> Otherwise, I tested the change without the comment and it seems to fix the issue we were seeing on Ubuntu 24.04 LTS.

@gjolly , could you please help review and test it again with the new updates?

proxy_agent_setup/src/linux/azure-proxy-agent.service
danpdraper: Including `After=sysinit.target` in the service configuration also leads to a cycle with `cloud-init`. The following is from a machine launched from an Ubuntu 24.04 LTS image containing a version of `azure-proxy-agent` whose service configuration includes all of `After=sysinit.target`, `Before=shutdown.target`, and `Conflicts=shutdown.target`:

```
Welcome to Ubuntu 24.04.3 LTS!
...
[    7.114416] systemd[1]: sysinit.target: Found ordering cycle on cloud-init.service/start
[    7.118263] systemd[1]: sysinit.target: Found dependency on systemd-networkd-wait-online.service/start
[    7.122182] systemd[1]: sysinit.target: Found dependency on systemd-networkd.service/start
[    7.125217] systemd[1]: sysinit.target: Found dependency on network-pre.target/start
[    7.127886] systemd[1]: sysinit.target: Found dependency on azure-proxy-agent.service/start
[    7.130841] systemd[1]: sysinit.target: Found dependency on sysinit.target/start
[    7.133641] systemd[1]: sysinit.target: Job cloud-init.service/start deleted to break ordering cycle starting with sysinit.target/start
[ SKIP ] Ordering cycle found, skipping cloud-init.service
```

danpdraper:
> > Otherwise, I tested the change without the comment and it seems to fix the issue we were seeing on Ubuntu 24.04 LTS.
> 
> @gjolly , could you please help review and test it again with the new updates?

@ZhidongPeng See https://github.com/Azure/GuestProxyAgent/pull/272#discussion_r2370319319 .

proxy_agent_setup/src/linux/azure-proxy-agent.service
```diff
@@ -1,8 +1,13 @@
...
 After=local-fs.target
 Before=network-pre.target
-Wants=network-pre.target
+# Baseline init + clean shutdown
+After=sysinit.target
```
ZhidongPeng: Including After=sysinit.target in the service configuration also leads to a cycle with cloud-init. The following is from a machine launched from a Noble image containing my patched version of azure-proxy-agent:
 
Welcome to Ubuntu 24.04.3 LTS!
...
[    7.114416] systemd[1]: sysinit.target: Found ordering cycle on cloud-init.service/start
[    7.118263] systemd[1]: sysinit.target: Found dependency on systemd-networkd-wait-online.service/start
[    7.122182] systemd[1]: sysinit.target: Found dependency on systemd-networkd.service/start
[    7.125217] systemd[1]: sysinit.target: Found dependency on network-pre.target/start
[    7.127886] systemd[1]: sysinit.target: Found dependency on azure-proxy-agent.service/start
[    7.130841] systemd[1]: sysinit.target: Found dependency on sysinit.target/start
[    7.133641] systemd[1]: sysinit.target: Job cloud-init.service/start deleted to break ordering cycle starting with sysinit.target/start
[ SKIP ] Ordering cycle found, skipping cloud-init.service

nmeyerhans (approved):

PR #284, opened 2025-10-07
Replace hmac_sha256 with native crypto crates

ZhidongPeng:
- added openssl in inux for hmac_sha256
- used windows-sys crate for hmac_sha256
- removed hmac_sha256 crate dependency
- static link openssl for linux musl target

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -124,6 +124,7 @@ guiddef
 handleapi
 hklm
 hlist
+HMAC
 hostga
 httpwg
 Iaa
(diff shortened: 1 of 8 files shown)


ZhidongPeng:
Since statically linked crypto is not normally recommended but in this case is better overusing an unapproved crypto library, the Crypto Board acknowledges that your team is aware of the potential issues you may be facing when using statically linked OpenSSL, and given that your team is attesting that the right steps will be followed.

Monitor OpenSSL CVEs: https://rustsec.org/packages/openssl.html

shahneerali (approved):

PR #294, opened 2025-12-01
set gpa service resource limits

ZhidongPeng:
1. Set the resource limits before launching the lower priority tasks.
2. Set the cpu limits for Linux GPA service and monitor memory usage.
3. Set the CPU and memory limits for Windows GPA service.
4. Removed the GPA service resource limits from its windows VM Extension

### DIFF
--- a/proxy_agent/src/provision.rs
+++ b/proxy_agent/src/provision.rs
@@ -314,7 +314,7 @@ pub async fn provision_timeup(
 }
 
 /// Set CPUQuota for azure-proxy-agent service
-fn set_cpu_quota() {
+fn set_resource_limits() {
     #[cfg(not(windows))]
     {
         // Set CPUQuota for azure-proxy-agent.service to 15% to limit the CPU usage for Linux azure-proxy-agent service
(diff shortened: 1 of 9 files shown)


proxy_agent/src/provision.rs
```diff
@@ -313,6 +313,31 @@ pub async fn provision_timeup(
...
 }
 
+fn set_cpu_quota() {
+    // Set CPUQuota for azure-proxy-agent.service to 15% to limit the CPU usage for Linux azure-proxy-agent service
+    // While windows does not need to set CPU quota as Windows GPA VMExtension already limit the CPU usage for GPA service
+    #[cfg(not(windows))]
+    {
+        match misc_helpers::execute_command(
+            "systemctl",
+            vec!["set-property", "azure-proxy-agent.service", "CPUQuota=15%"],
```
nagworld9: if you use Linux agent in the future, both conflict each other. So, make sure not to add this service in the resource governance.

nagworld9: I see you hard coded service name, if you already have service name in config or somewhere, better to use reuse to avoid mistakes in the future whenever the service name changes

nagworld9: do you run this cmd on every service start?

ZhidongPeng: yes, Linux GPA VM Extension will not have resource enforcement but only resource monitoring.

ZhidongPeng: yes

shahneerali (approved):

PR #301, opened 2026-01-05
Update redirect policy when provision finished

ZhidongPeng:
- Move `update the redirect policy maps` from `redirector` module to `provision` module. It will help not enable the host http intercept by default when provisioning `redirector` module.
- Removed the configuration setting for HostGAPlugin endpoint
- Break down KeyKeepr loop_poll function into smaller, more manageable sub-functions to improve readability and maintainability
- HostGAPlugin to use WireServer rules to match its mode.

### DIFF
--- a/proxy_agent/src/key_keeper.rs
+++ b/proxy_agent/src/key_keeper.rs
@@ -209,7 +209,7 @@ impl KeyKeeper {
             ));
         }
 
-        let mut start = Instant::now();
+        let mut provision_start_time = Instant::now();
         let mut redirect_policy_updated = false;
         loop {
             if !first_iteration {
@@ -233,7 +233,7 @@ impl KeyKeeper {
                     .await;
 
                 if reset_timer {
-                    start = Instant::now();
+                    provision_start_time = Instant::now();
                     provision_timeout = false;
                 }
 
@@ -243,7 +243,7 @@ impl KeyKeeper {
             }
             first_iteration = false;
 
-            self.handle_provision_timeout(&mut start, &mut provision_timeout)
+            self.handle_provision_timeout(&mut provision_start_time, &mut provision_timeout)
                 .await;
             started_event_threads = self.handle_event_threads_start(started_event_threads).await;
 
@@ -428,8 +428,8 @@ impl KeyKeeper {
     }
 
     /// Handle provision timeout logic
-    async fn handle_provision_timeout(&self, start: &mut Instant, provision_timeout: &mut bool) {
-        if !*provision_timeout && start.elapsed().as_millis() > PROVISION_TIMEOUT_IN_MILLISECONDS {
+    async fn handle_provision_timeout(&self, provision_start_time: &mut Instant, provision_timeout: &mut bool) {
+        if !*provision_timeout && provision_start_time.elapsed().as_millis() > PROVISION_TIMEOUT_IN_MILLISECONDS {
             provision::provision_timeout(
                 None,
                 self.provision_shared_state.clone(),
--- a/proxy_agent/src/key_keeper.rs
+++ b/proxy_agent/src/key_keeper.rs
@@ -619,6 +619,8 @@ impl KeyKeeper {
         Ok(())
     }
 
+    /// update the redirector/eBPF policy based on the secure channel status
+    /// it should be called when the secure channel state is changed
     async fn update_redirector_policy(&self, status: KeyStatus) -> bool {
         // update the redirector policy map
         if !redirector::update_wire_server_redirect_policy(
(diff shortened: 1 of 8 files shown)


shahneerali (dismissed):

proxy_agent/src/key_keeper.rs
```diff
@@ -220,383 +218,432 @@ impl KeyKeeper {
...
-                provision::provision_timeup(
-                    None,
-                    self.provision_shared_state.clone(),
-                    self.agent_status_shared_state.clone(),
-                )
-                .await;
-                provision_timeup = true;
+                if !continue_loop {
+                    continue;
+                }
```
huiminyan2017, line 242: Is this dead code? seems continue_loop always return true in handle_notification

ZhidongPeng: That's true for now, but it may return false in future.

proxy_agent/src/key_keeper.rs
```diff
@@ -220,383 +218,432 @@ impl KeyKeeper {
...
+        }
+    }
+
+    /// Handle notification and sleep logic
+    /// Returns (continue_loop, reset_timer)
+    async fn handle_notification(
+        &self,
+        notify: &tokio::sync::Notify,
+        sleep: Duration,
+        current_state: &str,
```
huiminyan2017, line 327: Took me sometime to understand this function, it's not easily readable, suggest to further breakdown it a bit to how the loop wakeup, and then handle wakeup, something like:

enum WakeReason {
    Notified,
    TimerElapsed,
}

async fn wait_for_wake(
    notify: &tokio::sync::Notify,
    sleep: Duration,
) -> WakeReason {
    tokio::select! {
        // if A notification arrives
        _ = notify.notified() => WakeReason::Notified,
        // if the sleep duration elapses
        _ = tokio::time::sleep(sleep) => WakeReason::TimerElapsed,
    }
}

async fn handle_wake(
    &self,
    wake_reason: WakeReason,
    current_state: &str,
) -> bool resetTimer {

    match wake_reason {
        WakeReason::TimerElapsed => false,

        WakeReason::Notified => {
            if current_state == DISABLE_STATE || current_state == UNKNOWN_STATE {
                provision::key_latch_ready_state_reset
                self.key_keeper_shared_state.update_current_secure_channel_state

                true
            } else {
                // report key has latched
                ...
                false
            }
        }
       
    }
}


async fn handle_notification(...) -> bool resetTimer
{
    let wake_reason = wait_for_wake(&notify, sleep).await;
    let reset_timer = self.handle_wake(wake_reason, &current_state).await;
}

proxy_agent/src/key_keeper.rs
```diff
@@ -220,383 +218,432 @@ impl KeyKeeper {
...
+                    // report key latched ready to try update the provision finished time_tick
+                    provision::key_latched(self.create_event_threads_shared_state()).await;
+                    let slept_time_in_millisec = time.elapsed().as_millis();
+                    let continue_sleep = sleep.as_millis() - slept_time_in_millisec;
+                    if continue_sleep > 0 {
+                        let continue_sleep = Duration::from_millis(continue_sleep as u64);
+                        let message = format!("poll_secure_channel_status task notified but secure channel state is '{current_state}', continue with sleep wait for {continue_sleep:?}.");
+                        logger::write_warning(message);
+                        tokio::time::sleep(continue_sleep).await;
+                    }
```
huiminyan2017: Why this sleep logic needed? if receiving a notification when the state is already enabled, you want to reduce the rate to loop again immediately? 
can you see if you can move the sleep logic a bit upward to the caller, it feels tangled to have sleep logic here. Add some comment if cannot improve

proxy_agent/src/key_keeper.rs
```diff
@@ -220,383 +218,432 @@ impl KeyKeeper {
...
-                        if let Err(e) = self
-                            .access_control_shared_state
-                            .set_imds_rules(status.get_imds_rules())
-                            .await
-                        {
-                            logger::write_error(format!("Failed to set imds rules: {e}"));
-                        }
-                        access_control_rules_changed = true;
+    /// Handle provision timeout logic
+    async fn handle_provision_timeup(&self, start: &mut Instant, provision_timeup: bool) -> bool {
```
huiminyan2017: Why not make provision_timeup mutable and modify it  in place, should be more readable at least

huiminyan2017: Also change all timeup to timeout pls

proxy_agent/src/key_keeper.rs
ZhidongPeng: Thanks

proxy_agent/src/key_keeper.rs
ZhidongPeng: Yes, it is needed to avoid frequent query/notify to wake up our pull task. Let me add comments to address the concerns.

huiminyan2017 (approved):

PR #307, opened 2026-01-15
Add EventReaderLimits into telemetry event reader mod

ZhidongPeng:
event_reader.rs major updates:

- the event json file name must be numeric characters
- add event file size limit, if the event json file size exceeds the limit, file will be deleted immediately
- add max events number limit per round, if the processed event number exceeds the limits, other files will not be skipped at current round.
- add `version` to override version and event_name of TelemetryEvent


proxy_agent_shared/src/telemetry/event_reader.rs
```diff
@@ -59,16 +59,52 @@ impl VmMetaData {
     }
 }
 
+/// Configuration for limiting EventReader behavior
+#[derive(Default, Clone)]
+pub struct EventReaderLimits {
+    pub max_events_per_round: Option<usize>,
+    pub max_event_file_size_bytes: Option<u64>,
+    pub version: Option<String>,
```
bingyuyin, line 67: note to myself: what's the reason that 'version' binded together with those events limits?
given that version is per extension and those events limit is shared across different extension versions  

=========>
that's GA version not extension version

proxy_agent_shared/src/telemetry/event_reader.rs
```diff
@@ -240,6 +302,42 @@ impl EventReader {
...
         for file in files {
+            if let Some(max_events) = self.limits.max_events_per_round {
+                if num_events_logged >= max_events {
+                    logger_manager::write_warn(format!(
+                                "EventReader:: Reached the max number of events to be read per round: {}. Stop processing file {} this round.",
+                                max_events,
+                                file.display()
+                            ));
+                    // do not delete this event json file, will try process it at next round
+                    continue;
```
bingyuyin: will use 'break' to quick exit the loop better then 'continue'?

proxy_agent_shared/src/telemetry/event_reader.rs
```diff
@@ -145,24 +203,28 @@ impl EventReader {
...
     async fn process_events(
         &self,
         wire_server_client: &WireServerClient,
         vm_meta_data: &VmMetaData,
     ) -> usize {
         let event_count: usize;
-        // get all .json event files in the directory
-        match misc_helpers::search_files(&self.dir_path, r"^(.*\.json)$") {
+        // get all [0-9]+.json event filenames with numbers in the directory
+        match misc_helpers::search_files(&self.dir_path, r"^[0-9]+\.json$") {
```
bingyuyin, line 227: any impact on existing flow with changing the regex matching here?

proxy_agent_shared/src/telemetry/event_reader.rs
```diff
@@ -59,16 +59,52 @@ impl VmMetaData {
     }
 }
 
+/// Configuration for limiting EventReader behavior
+#[derive(Default, Clone)]
+pub struct EventReaderLimits {
+    pub max_events_per_round: Option<usize>,
+    pub max_event_file_size_bytes: Option<u64>,
```
bingyuyin, line 66: plan to add single event size limit?

proxy_agent_shared/src/telemetry/event_reader.rs
ZhidongPeng: no, currently they all numbers and Unit test covered how GPA currently generate the event json file name

proxy_agent_shared/src/telemetry/event_reader.rs
ZhidongPeng: We donot have it in production yet.

proxy_agent_shared/src/telemetry/event_reader.rs
ZhidongPeng: good catch, break could exit the loop earlier instead of continuously max_events check.

bingyuyin (approved):

PR #313, opened 2026-02-04
Support report extension status event and send out

ZhidongPeng:
- Extend `event_logger.rs` and `telemetry.rs` to able write extension status event to `events` folder with different file name pattern
- Extend `event_reader.rs` to read/process extension status events
- Added more tests to cover extension status events and mixed types of events.
- Added `ExtensionEvent(TelemetryExtensionEventsEvent)` into `enum TelemetryEvent`
- Fixed the missing fields from `TelemetryGenericLogsEvent::to_xml_event`
- GPA to start 2 async tasks to read generic events and extension status events
- GPA status loop to use `report_extension_status_event`
- Replace `misc_helper::get_current_version` with `current_info::get_current_exe_version` at logging and events
- Fixed the test server_mock to use an ephemeral port if the port is already in use, also returns the actual-binding port to the caller

### DIFF
--- a/proxy_agent/src/provision.rs
+++ b/proxy_agent/src/provision.rs
@@ -382,6 +382,19 @@ pub async fn start_event_threads(event_threads_shared_state: EventThreadsSharedS
         }
     });
     tokio::spawn({
+        let event_reader = EventReader::new(
+            config::get_events_dir(),
+            event_threads_shared_state.common_state.clone(),
+            "ProxyAgent".to_string(),
+            "MicrosoftAzureGuestProxyAgent".to_string(),
+        );
+        async move {
+            event_reader
+                .start_extension_status_event_processor(true, Some(Duration::from_secs(60)))
+                .await;
+        }
+    });
+    tokio::spawn({
         let event_sender = EventSender::new(event_threads_shared_state.common_state.clone());
         async move {
             event_sender.start(None, None).await;
--- a/proxy_agent/src/proxy/proxy_server.rs
+++ b/proxy_agent/src/proxy/proxy_server.rs
@@ -564,7 +564,7 @@ impl ProxyServer {
 
         if http_connection_context.should_skip_sig() {
             http_connection_context.log(
-                LoggerLevel::Info,
+                LoggerLevel::Trace,
                 format!(
                     "Skip compute signature for the request for {} {}",
                     http_connection_context.method, http_connection_context.url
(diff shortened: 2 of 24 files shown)


shahneerali (dismissed):

proxy_agent/src/proxy_agent_status.rs
```diff
@@ -140,12 +144,17 @@ impl ProxyAgentStatusTask {
...
                     Err(e) => format!("Error serializing proxy agent status: {e}"),
                 };
-                event_logger::write_event(
-                    LoggerLevel::Info,
-                    status,
-                    "loop_status",
-                    "proxy_agent_status",
-                    logger::AGENT_LOGGER_KEY,
+                // private build tests report_extension_status_event
+                event_logger::report_extension_status_event(
```
bingyuyin, line 147: private build test only, need to revert or plan to change here?

proxy_agent/src/proxy/proxy_server.rs
```diff
@@ -564,7 +564,7 @@ impl ProxyServer {
 
         if http_connection_context.should_skip_sig() {
             http_connection_context.log(
-                LoggerLevel::Info,
+                LoggerLevel::Trace,
```
bingyuyin, line 567: to myself: trace and info difference?

ZhidongPeng: trace will not send to kusto

proxy_agent/src/proxy_agent_status.rs
ZhidongPeng: good catch, it is not a test now. I want to have GAExtensionStatus event in GPA too.

proxy_agent/src/provision.rs
```diff
@@ -381,6 +381,19 @@ pub async fn start_event_threads(event_threads_shared_state: EventThreadsSharedS
                 .await;
         }
     });
+    tokio::spawn({
+        let event_reader = EventReader::new(
+            config::get_events_dir(),
```
bingyuyin, line 386: to myself: how does the events under GPA events folder be sent before?

ZhidongPeng: yes, it is the same as event_logger

ZhidongPeng: **before**: event_reader reads the generic events and send out to wireserver;
**now**: 
- event_reader reads the generic and extension status events and push to memory queue and fire `notify` event
- event_sender responds to `notify` event and read the events from the memory queue and sends to WireServer.

bingyuyin: OK, I see, event_reader.start is for generic event, and start_extension_status_event_processor is for extension event,

bingyuyin (approved):

PR #317, opened 2026-02-10
regex perf improvements

ZhidongPeng:
- `Regex::New` is performance-sensitive - use `LazyLock` to compile it only once as a static regex and reuse it for subsequent calls
- `search_files` should not create a new regex instance each time, and the caller should cache the regex if it is used frequently.
- Config setting value `get_xxx` in Config invoked `resolve_env_variables` at each call - resolve all the config values at new when first read & parse our config file.
- `regex.replace_all`'s time-complexity is` O(m * n^2)` in `redact_secrets` - first check for indicators, just simple substrings to check for before running the more expensive regexes.

### DIFF
--- a/proxy_agent/src/common/config.rs
+++ b/proxy_agent/src/common/config.rs
@@ -122,39 +122,44 @@ impl Default for Config {
             config_file_full_path = misc_helpers::get_current_exe_dir();
             config_file_full_path.push(CONFIG_FILE_NAME);
         }
+
         Config::from_json_file(config_file_full_path)
     }
 }
 
 impl Config {
+    /// Load config from a specific JSON file path
     pub fn from_json_file(file_path: PathBuf) -> Self {
-        misc_helpers::json_read_from_file::<Config>(&file_path).unwrap_or_else(|_| {
-            panic!(
-                "Error in reading Config from Json file: {}",
-                misc_helpers::path_to_string(&file_path)
-            )
-        })
+        let mut config =
+            misc_helpers::json_read_from_file::<Config>(&file_path).unwrap_or_else(|_| {
+                panic!(
+                    "Error in reading Config from Json file: {}",
+                    misc_helpers::path_to_string(&file_path)
+                )
+            });
+        config.resolve_env_variables();
+        config
+    }
+
+    /// Resolve environment variables in path fields once during construction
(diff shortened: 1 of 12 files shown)


proxy_agent/src/common/config.rs
```diff
@@ -122,39 +122,44 @@ impl Default for Config {
...
+    /// This allows us to keep the rest of the code simple without worrying about env vars,
+    /// and also ensures that we only pay the cost of resolving env vars once at startup rather than on every access.
+    fn resolve_env_variables(&mut self) {
+        self.logFolder = misc_helpers::resolve_env_variables(&self.logFolder);
+        self.eventFolder = misc_helpers::resolve_env_variables(&self.eventFolder);
+        self.latchKeyFolder = misc_helpers::resolve_env_variables(&self.latchKeyFolder);
     }
 
     pub fn get_log_folder(&self) -> String {
-        match misc_helpers::resolve_env_variables(&self.logFolder) {
```
shahneerali, line 140: why do we remove match?   misc_helpers::resolve_env_variables doesn't contain match, it returns prev val

shahneerali (changes requested):

proxy_agent/src/common/config.rs
ZhidongPeng: since I changed `resolve_env_variables`, there is no possible to throw `error`

ZhidongPeng: The original `error` already ignored in the new code 
```
/// Static regex for matching environment variables like %VAR%
/// expect() only panics on failure to compile the regex, which should not happen since the pattern is constant and valid
/// This is the idiomatic Rust pattern for creating a static regex that is compiled once and reused, ensuring thread safety and performance
/// Remark: Regex::new is performance-sensitive, so we use LazyLock to compile it only once and reuse it for subsequent calls to resolve_env_variables, which can be called frequently.
///     This avoids the overhead of compiling the regex on every call, improving performance while ensuring thread safety.
static ENV_VAR_REGEX: std::sync::LazyLock<Regex> =
    std::sync::LazyLock::new(|| Regex::new(r"%(\w+)%").expect("Invalid env var regex pattern"));
```

shahneerali (approved):

PR #323, opened 2026-02-20
Fix: Privilege match must be case insensitive.

ZhidongPeng:
- Fix Privilege path match case insensitive issue
- Updated UT to verify the path case insensitive match.
- Optimize: store lowercased versions of path, query parameter keys, and query parameter values in ComputedAuthorizationItem at construction time
- Optimize: lowercase the request URL path once before iterating over privileges

### DIFF
--- a/proxy_agent/src/key_keeper/key.rs
+++ b/proxy_agent/src/key_keeper/key.rs
@@ -213,12 +213,20 @@ impl Clone for Privilege {
 }
 
 impl Privilege {
-    pub fn is_match(&self, logger: &mut ConnectionLogger, request_url: &Uri) -> bool {
+    /// Note: `self.path` and `self.queryParameters` keys/values are expected to be
+    /// pre-lowercased (done in `ComputedAuthorizationItem::from_authorization_item`).
+    /// `lowered_request_path` should be `request_url.path().to_lowercase()`, hoisted by the caller.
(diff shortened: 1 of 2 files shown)


proxy_agent/src/key_keeper/key.rs
```diff
@@ -213,12 +213,20 @@ impl Clone for Privilege {
 }
 
 impl Privilege {
-    pub fn is_match(&self, logger: &mut ConnectionLogger, request_url: &Uri) -> bool {
+    /// Note: `self.path` and `self.queryParameters` keys/values are expected to be
+    /// pre-lowercased (done in `ComputedAuthorizationItem::from_authorization_item`).
+    /// `lowered_request_path` should be `request_url.path().to_lowercase()`, hoisted by the caller.
```
bingyuyin, line 218: if the   `lowered_request_path` always be `request_url.path().to_lowercase()`, why not remove the lowered_request_path as an input parameter, just got from request_url inside the function?

bingyuyin (approved):
approve with sugguestion

proxy_agent/src/key_keeper/key.rs
ZhidongPeng: this is one of my Optimize: lowercase the request URL path **once before iterating** over privileges

PR #329, opened 2026-03-18
Add local file-based access-control rule support.

ZhidongPeng:
- Added base64 = "0.22" dependency
- Introduces a new local_rules module that 
> parses base64-encoded rule-id descriptors, 
> merges host-delivered and customer-managed rules with fail-closed behavior on parse errors, 
> tracks file state across polls, and integrates with key_keeper 
- Added rules_dir: PathBuf field to KeyKeeper struct and refactored update_access_control_rules() to accept state tracker, call resolve_effective_rules() for WireServer/IMDS/HostGA, and handle local-rule-merged effective rules

### DIFF
--- a/proxy_agent/src/key_keeper.rs
+++ b/proxy_agent/src/key_keeper.rs
@@ -29,8 +29,7 @@ pub mod local_rules;
 
 use self::key::Key;
 use self::local_rules::{
-    get_rules_dir_from_key_dir, resolve_effective_rules, LocalRuleStateTracker,
-    LocalRuleTarget,
+    get_rules_dir_from_key_dir, resolve_effective_rules, LocalRuleStateTracker, LocalRuleTarget,
 };
 use crate::common::error::{Error, KeyErrorType};
 use crate::common::result::Result;
@@ -579,7 +578,11 @@ impl KeyKeeper {
                 imds: imds_rules.clone(),
                 hostga: hostga_rules.clone(),
             });
-            if let (Ok(computed_wireserver_rules), Ok(computed_imds_rules), Ok(computed_hostga_rules)) = (
+            if let (
+                Ok(computed_wireserver_rules),
+                Ok(computed_imds_rules),
+                Ok(computed_hostga_rules),
+            ) = (
                 self.access_control_shared_state
                     .get_wireserver_rules()
                     .await,
@@ -601,7 +604,7 @@ impl KeyKeeper {
         access_control_rules_changed
     }
 
-    /// Ensure the directory exists and has secure ACLs. 
+    /// Ensure the directory exists and has secure ACLs.
     /// If the directory does not exist, it will be created.
     fn ensure_secure_directory(&self, dir: &Path, dir_kind: &str) {
         if let Err(e) = misc_helpers::try_create_folder(dir) {
--- a/proxy_agent/src/key_keeper.rs
+++ b/proxy_agent/src/key_keeper.rs
@@ -536,7 +536,6 @@ impl KeyKeeper {
             imds_rule_id_changed,
         )
         .await;
-        let hostga_rules = wireserver_rules.clone();
 
         if wireserver_rule_id_changed || wireserver_local_state_changed {
             if let Err(e) = self
@@ -560,7 +559,9 @@ impl KeyKeeper {
             access_control_rules_changed = true;
         }
 
-        if hostga_rule_id_changed || wireserver_rule_id_changed || wireserver_local_state_changed {
+        // HostGA rules only come from server and do not have local rules, so only update when rule id changed
+        let hostga_rules = status.get_hostga_rules();
+        if hostga_rule_id_changed {
             if let Err(e) = self
                 .access_control_shared_state
                 .set_hostga_rules(hostga_rules.clone())
--- a/proxy_agent/src/key_keeper.rs
+++ b/proxy_agent/src/key_keeper.rs
@@ -25,12 +25,17 @@
 //! ```
 
 pub mod key;
+pub mod local_rules;
 
 use self::key::Key;
+use self::local_rules::{
+    get_rules_dir_from_key_dir, resolve_effective_rules, LocalRuleStateTracker,
+    LocalRuleTarget,
+};
 use crate::common::error::{Error, KeyErrorType};
 use crate::common::result::Result;
 use crate::common::{constants, helpers, logger};
-use crate::key_keeper::key::KeyStatus;
+use crate::key_keeper::key::{AuthorizationRules, KeyStatus};
 use crate::provision;
 use crate::proxy::authorization_rules::{AuthorizationRulesForLogging, ComputedAuthorizationRules};
 use crate::shared_state::access_control_wrapper::AccessControlSharedState;
@@ -70,6 +75,8 @@ pub struct KeyKeeper {
     port: u16,
     /// key_dir: the folder to save the key details
     key_dir: PathBuf,
+    /// rules_dir: the folder to save customer-managed local access control rules
+    rules_dir: PathBuf,
     /// status_dir: the folder to log the access control rule details
     status_dir: PathBuf,
     /// interval: the interval to poll the secure channel status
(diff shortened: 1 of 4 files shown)


proxy_agent/src/key_keeper.rs
```diff
@@ -531,40 +508,93 @@ impl KeyKeeper {
...
+        .await;
+        let (imds_rules, imds_local_state_changed) = resolve_effective_rules(
+            &self.rules_dir,
+            status.get_imds_rules(),
+            LocalRuleTarget::Imds,
+            &mut local_rule_state_tracker.imds,
+            imds_rule_id_changed,
+        )
+        .await;
+        let hostga_rules = wireserver_rules.clone();
```
ZhidongPeng: **[Potential Regression]** `hostga_rules` is now cloned from `wireserver_rules` (which includes local-rule merges). Before this PR, HostGA used its own independent rules from `status.get_hostga_rules()`.

This is a behavioral change: HostGA rules will now always mirror WireServer's effective rules, ignoring whatever the host delivers for HostGA specifically. The HostGA rule-id is still tracked for changes (line 508-519) but no longer governs the actual rules applied.

Was this intentional? If so, please add a comment explaining why HostGA should mirror WireServer. If not, consider adding a `LocalRuleTarget::HostGA` variant (or at minimum `let hostga_rules = status.get_hostga_rules();` for the non-local-rule path).

proxy_agent/src/key_keeper/local_rules.rs
```diff
@@ -0,0 +1,707 @@
...
+        || file_state_changed
+        || previous_parse_failed;
+
+    if !needs_refresh {
+        return (tracker.effective_rules.clone(), false);
+    }
+
+    if matches!(current_file_state, LocalRuleFileState::Missing) {
+        tracker.parse_failed = false;
+        tracker.effective_rules = None;
```
ZhidongPeng: **[Security Concern]** When `use_local_file_rules` is enabled but the file is missing, `effective_rules` becomes `None`. Depending on how the proxy treats `None` rules, this could mean **no access control is enforced at all**.

Compare with the parse-error path (line 372) which properly fail-closes with `defaultAccess: deny` and no rules. Consider applying similar fail-closed behavior here, or at the very least documenting why `None` is safe for the file-missing case. For example:

`
ust
// File missing with local rules enabled: fail-closed
tracker.effective_rules = build_fail_closed_rules(normalized_remote_rules, &descriptor);
`

proxy_agent/src/key_keeper/local_rules.rs
```diff
@@ -0,0 +1,707 @@
...
+        defaultAccess: "deny".to_string(),
+        mode: "enforce".to_string(),
+        id: descriptor.logical_id.clone(),
+        rules: None,
+    });
+
+    if !descriptor.logical_id.is_empty() {
+        rules.id = descriptor.logical_id.clone();
+    }
+    rules.defaultAccess = "deny".to_string();
```
ZhidongPeng, line 305: **[Question]** `build_fail_closed_rules` forces `defaultAccess = ""deny""` and clears `rules`, but preserves the original `mode` from the remote `AuthorizationItem`. If the remote mode is `""disabled""`, the result would be `mode=""disabled"", defaultAccess=""deny"", rules=None`.

Is that still effectively fail-closed? If `mode=""disabled""` causes the proxy to skip access-control enforcement entirely, the `deny` default might never take effect. Consider also forcing `mode = ""enforce""` here for a true fail-closed posture.

proxy_agent/src/key_keeper/local_rules.rs
```diff
@@ -0,0 +1,707 @@
...
+
+/// Get the state of the local rule file - whether it is present or missing, and if present, its last modified time.
+pub(crate) fn get_local_rule_file_state(file_path: &Path) -> LocalRuleFileState {
+    match fs::metadata(file_path) {
+        Ok(metadata) => match metadata.modified() {
+            Ok(modified) => LocalRuleFileState::Present(modified),
+            Err(_) => LocalRuleFileState::Present(SystemTime::UNIX_EPOCH),
+        },
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => LocalRuleFileState::Missing,
+        Err(_) => LocalRuleFileState::Missing,
```
ZhidongPeng: **[Medium]** This catch-all treats permission errors (`PermissionDenied`, etc.) the same as `NotFound`. A permission error on a security-sensitive file is arguably a different scenario ??? it could indicate misconfigured ACLs or a privilege escalation attempt.

Consider logging a warning for non-`NotFound` errors, and/or introducing a separate `LocalRuleFileState::Error` variant so callers can distinguish this case and fail-closed:

`
ust
Err(e) if e.kind() == std::io::ErrorKind::NotFound => LocalRuleFileState::Missing,
Err(e) => {
    logger::write_warning(format!(
        ""Unexpected error reading local rules file metadata: {e}""
    ));
    LocalRuleFileState::Missing // or a new Error variant
}
`

proxy_agent/src/key_keeper/local_rules.rs
```diff
@@ -0,0 +1,707 @@
...
+        return (
+            normalized_remote_rules,
+            use_local_file_rules_changed || previous_parse_failed,
+        );
+    }
+
+    let local_rules_file = rules_dir.join(target.file_name());
+    let current_file_state = get_local_rule_file_state(&local_rules_file);
+    let file_state_changed = tracker.file_state != current_file_state;
+
```
ZhidongPeng, line 623: **[Minor]** There is a small TOCTOU window between checking `get_local_rule_file_state` (metadata/mtime) and actually reading the file later (line 351). The file could be modified or deleted in between. The retry logic in `read_local_rules_file` partly mitigates this, but in the worst case:

1. `get_local_rule_file_state` sees `Present(T1)`
2. File is deleted
3. `read_local_rules_file` fails all retries ??? fail-closed (correct behavior)

So the retry + fail-closed combination makes this safe in practice. Just wanted to flag it for awareness.

proxy_agent/src/key_keeper/local_rules.rs
```diff
@@ -0,0 +1,707 @@
...
+        use_local_file_rules: false,
+    }
+}
+
+pub(crate) fn normalize_authorization_item(
+    authorization_item: Option<AuthorizationItem>,
+    descriptor: &RuleIdDescriptor,
+) -> Option<AuthorizationItem> {
+    authorization_item.map(|mut item| {
+        if !descriptor.logical_id.is_empty() {
```
ZhidongPeng: **[Nit]** If the base64-decoded JSON has `id: """"` (empty string), `normalize_authorization_item` skips the id override (because `descriptor.logical_id.is_empty()`), leaving the original base64-encoded string as the item's `id`. This could produce confusing log output. Consider defaulting to the raw rule id as a fallback in `parse_rule_id_descriptor` when the decoded `id` is empty.

proxy_agent/src/key_keeper.rs
```diff
@@ -531,40 +508,93 @@ impl KeyKeeper {
...
+                .set_imds_rules(imds_rules.clone())
+                .await
+            {
+                logger::write_error(format!("Failed to set imds rules: {e}"));
+            }
+            access_control_rules_changed = true;
+        }
+
+        if hostga_rule_id_changed || wireserver_rule_id_changed || wireserver_local_state_changed {
+            if let Err(e) = self
```
ZhidongPeng, line 582: **[Nit]** Since `hostga_rules` is now always `wireserver_rules.clone()`, the `hostga_rule_id_changed` flag only triggers an extra `set_hostga_rules` call without actually using HostGA-specific rules. The HostGA rule-id tracking (lines 508-519) is still needed to detect change for the update trigger, but a brief comment here explaining why HostGA depends on wireserver would help maintainability.

ZhidongPeng (commented):
## PR Review ??? Add local file-based access-control rule support

Thanks for the PR. Overall the design is clean: base64-encoded rule-id descriptors, fail-closed on parse errors, and file-state tracking across polls are all solid patterns.

Below are inline comments ranging from a potential behavioral regression to minor nits. Please take a look.

proxy_agent/src/key_keeper/local_rules.rs
ZhidongPeng: added with new Error variant `LocalRuleFileState::Error(String)`

proxy_agent/src/key_keeper/local_rules.rs
ZhidongPeng: if mode="disabled" they will be no proxied requests via eBPF.

srikrishnaveturi (approved):

PR #335, opened 2026-04-15
GPA service to use host-date-time for signed http requests

ZhidongPeng:
When GPA proxy server constructs the http requests to Host endpoints, it must add new http request header `x-ms-azure-host-date`. 
Currently, it uses local utc time in RFC1123 format, while if the VM is just provisioned and the local time clock not synced yet, the time may not pass the validation at host side. 
To fix this issue, GPA needs to sync and cache the host time at `GET /secure-channel/status` API every 15 minutes. and then use the synced host-time to http request header `x-ms-azure-host-date`. 
When reads the host time from http response header, it first uses the custom date header `x-ms-azure-host-date`, if it is not available, fallbacks to standard http data header `date`.

### DIFF
--- a/proxy_agent/src/key_keeper/key.rs
+++ b/proxy_agent/src/key_keeper/key.rs
@@ -36,9 +36,9 @@ use hyper::Uri;
 use proxy_agent_shared::hyper_client;
 use proxy_agent_shared::logger::LoggerLevel;
 use serde_derive::{Deserialize, Serialize};
-use std::ffi::OsString;
 use std::fmt::{Display, Formatter};
 use std::{collections::HashMap, path::PathBuf};
+use std::{ffi::OsString, time::Duration};
 
 const AUDIT_MODE: &str = "audit";
 const ENFORCE_MODE: &str = "enforce";
@@ -729,7 +729,11 @@ impl Display for KeyAction {
 
 const STATUS_URL: &str = "/secure-channel/status";
 const KEY_URL: &str = "/secure-channel/key";
+const HOST_DATE_TIME_SYNC_MAX_AGE: Duration = Duration::from_secs(60);
 
+/// Get the current status of the key from the secure channel.
+/// This function will perform a GET request to the secure channel status endpoint.
+/// If the host time sync is stale, it will use `get_and_sync_host_time` to update the host time.
 pub async fn get_status(host: &str, port: u16) -> Result<KeyStatus> {
     let endpoint = hyper_client::HostEndpoint::new(host, port, STATUS_URL);
     let mut headers = HashMap::new();
@@ -737,8 +741,22 @@ pub async fn get_status(host: &str, port: u16) -> Result<KeyStatus> {
         hyper_client::METADATA_HEADER.to_string(),
         "True ".to_string(),
     );
+
     let status: KeyStatus =
-        hyper_client::get(&endpoint, &headers, None, None, logger::write_warning).await?;
+        if proxy_agent_shared::misc_helpers::host_time_sync_is_stale(HOST_DATE_TIME_SYNC_MAX_AGE) {
+            hyper_client::get_and_sync_host_time(
+                &endpoint,
+                &headers,
+                None,
+                None,
+                logger::write_warning,
+            )
+            .await?
+            .0
+        } else {
+            hyper_client::get(&endpoint, &headers, None, None, logger::write_warning).await?
+        };
+
     status.validate()?;
 
     Ok(status)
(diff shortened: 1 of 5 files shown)


proxy_agent/src/key_keeper/key.rs
```diff
@@ -729,16 +729,35 @@ impl Display for KeyAction {
 
 const STATUS_URL: &str = "/secure-channel/status";
 const KEY_URL: &str = "/secure-channel/key";
+const HOST_DATE_TIME_SYNC_MAX_AGE: Duration = Duration::from_secs(60 * 15);
```
LittleBoxOfSunshine: Consider using the term "drift" instead of sync for clarity

proxy_agent/src/key_keeper/key.rs
```diff
@@ -729,16 +729,35 @@ impl Display for KeyAction {
...
+                &headers,
+                None,
+                None,
+                logger::write_warning,
+            )
+            .await?;
+            logger::write(format!("Host time synced: {host_time_synced}",));
+            key_status
+        } else {
+            hyper_client::get(&endpoint, &headers, None, None, logger::write_warning).await?
```
LittleBoxOfSunshine: It might be simpler to unconditionally call this fn, then check if stale, and if stale do the sync. Having the if else here means we have the duplication and forked logic potential

proxy_agent_shared/src/misc_helpers.rs
```diff
@@ -63,11 +73,56 @@ pub fn get_date_time_string() -> String {
...
+    *state = Some(HostTimeSyncState {
+        synced_host_utc: parsed_host_utc.assume_utc(),
+        synced_instant: Instant::now(),
+    });
+    true
+}
+
+/// Returns true when current host-time sync state is older than `max_age`.
+/// If there is no host-time sync state yet, this returns true.
+pub fn host_time_sync_is_stale(max_age: std::time::Duration) -> bool {
```
LittleBoxOfSunshine, line 111: Passing in the limit this way gives the opportunity for the wrong value to be used. Consider either taking no arguments, or inverting it. Have the constant be referenced directly, and then pass in the state you want to check for drift

Also would framing this as that the clock drift is too high be better naming? e.g. `clock_drift_exceeds_host_tolerance`?

LittleBoxOfSunshine (dismissed):

proxy_agent_shared/src/misc_helpers.rs
```diff
@@ -823,4 +879,32 @@ mod tests {
...
+            !super::host_time_sync_is_stale(Duration::from_secs(3600)),
+            "Sync state should not be stale right after update"
+        );
+
+        std::thread::sleep(Duration::from_millis(1));
+        assert!(
+            super::host_time_sync_is_stale(Duration::from_millis(0)),
+            "Sync state should be stale when max_age is zero"
+        );
+    }
```
srikrishnaveturi, line 916: This tests sets the static variable `HOST_TIME_SYNC_STATE`, should this be reset at the end of the test?

proxy_agent_shared/src/hyper_client.rs
```diff
@@ -142,6 +142,59 @@ where
     read_response_body(response).await
 }
 
+/// Performs a GET request and updates the shared host-time sync state from
+/// `x-ms-azure-host-date` response header if available and parseable.
```
srikrishnaveturi: also need to document the fallback behavior

proxy_agent_shared/src/hyper_client.rs
```diff
@@ -142,6 +142,59 @@ where
...
+///
+/// Returns `Ok(true)` when sync state was updated, `Ok(false)` when header is
+/// missing/invalid, and `Err` on transport/server status failure.
+pub async fn get_and_sync_host_time<T, F>(
```
srikrishnaveturi: can we add some unit testing for both methods get_and_sync_host_time and sync_host_time_from_headers

proxy_agent_shared/src/misc_helpers.rs
ZhidongPeng: it is the shared crate; this parameter gives the caller an option to decide how frequently they want to.

proxy_agent/src/key_keeper/key.rs
ZhidongPeng: yeah, removed this odd `fn hyper_client::get_and_sync_host_time`

proxy_agent_shared/src/misc_helpers.rs
ZhidongPeng: yeah, it is good effort to reset it to None.

proxy_agent_shared/src/hyper_client.rs
ZhidongPeng: added ut for `sync_host_time_from_headers`

srikrishnaveturi (approved):

PR #346, opened 2026-05-13
Set Linux file‑mode for more restrictive permissions

ZhidongPeng:


proxy_agent/src/proxy_agent_status.rs
```diff
@@ -292,6 +292,16 @@ impl ProxyAgentStatusTask {
     async fn write_aggregate_status_to_file(&self, status: GuestProxyAgentAggregateStatus) {
         let full_file_path = self.status_dir.join("status.json");
         if let Err(e) = misc_helpers::json_write_to_file_async(&status, &full_file_path).await {
+            #[cfg(not(windows))]
+            {
+                proxy_agent_shared::linux::set_file_permissions(&full_file_path, 0o640)
```
srikrishnaveturi, line 302: Why are we setting permissions when the write failed? Should we also set the permissions inside the else block just under this?

ZhidongPeng: it is bug, we need set file permissions when write success intead.

srikrishnaveturi (approved):

PR #347, opened 2026-05-14
GPA proxy_server to have an explicit http-method allow-list

ZhidongPeng:


proxy_agent/src/proxy/proxy_server.rs
```diff
@@ -878,6 +905,48 @@ impl ProxyServer {
         response
     }
 
+    /// Methods that GPA's HTTP/1.1 forward-proxy will accept and try to route
+    /// to a fabric endpoint. Anything else (CONNECT, TRACE, made-up verbs)
+    /// is rejected with 405 by `handle_new_http_request` BEFORE any
+    /// authorization or upstream lookup happens.
+    const ALLOWED_METHODS: &'static [Method] = &[
```
srikrishnaveturi, line 912: High level question - Are there any other methods that customers could be using that should be allowed but are not in this list? Can we log the method type as part of the audit logs so customers understand if they are using anything else?

srikrishnaveturi (approved):

proxy_agent/src/proxy/proxy_server.rs
```diff
@@ -376,6 +377,32 @@ impl ProxyServer {
...
+        // through unsupported methods, and respond with an explicit 405 + Allow
+        // (RFC 7231 §6.5.5) so clients and pen-test harnesses see a real status
+        // line instead of a silent RST.
+        if !Self::is_method_allowed(&http_connection_context.method) {
+            let bad_method = http_connection_context.method.to_string();
+            self.log_connection_summary(
+                &mut http_connection_context,
+                StatusCode::METHOD_NOT_ALLOWED,
+                false,
+                format!("Rejected unsupported HTTP method '{bad_method}'"),
```
srikrishnaveturi, line 400: Do we have public docs on what methods are allowed? Or else we should add all the allowed types here

proxy_agent/src/proxy/proxy_server.rs
ZhidongPeng: I asked the review from Host side. Let's start with this list and we could adjust them if Host has any feedback.

proxy_agent/src/proxy/proxy_server.rs
ZhidongPeng: So far, all the HTTP methods are defined in IMDS public documentation https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=linux#http-verbs. While WS endpoints are internal.

PR #351, opened 2026-06-02
Add pen-test design and automation scripts

ZhidongPeng:


pentest/linux/phase4b_local_rules/__pycache__/run.cpython-312.pyc
srikrishnaveturi, line 1: Is this file necessary? Seems like local build binary

pentest/linux/phase4_rules_fuzz/url_diff.py
```diff
@@ -0,0 +1,85 @@
...
+    for name, path in VARIANTS.items():
+        status, head = send(path)
+        diff = "DIFF" if status != baseline_status else "same"
+        if diff == "DIFF":
+            diff_count += 1
+        rows.append((name, status, diff, urllib.parse.quote(path, safe="/?&=:%")[:120], head[:60]))
+        print(f"  {name:<18} status={status:<4} {diff}")
+
+    with open(args.out, "a", encoding="utf-8") as fh:
+        ts = datetime.datetime.utcnow().isoformat(timespec="seconds") + "Z"
```
srikrishnaveturi: This is currently deprecated, can we move to using `datetime.datetime.now()`?

pentest/linux/phase3_authn_authz/run.sh
```diff
@@ -0,0 +1,174 @@
...
+  record C1 INFO "skipped (need sudo + nobody account)"
+fi
+
+# C7 — alternate IP forms (hex, decimal) must not bypass the eBPF redirect /
+# AuthZ. Compare each alt's HTTP status AND response-body sha256 against the
+# canonical form captured above. Identical → eBPF treated them the same →
+# PASS. Different → potential SSRF-filter bypass at the network layer → FAIL.
+for alt in "http://0xa9fea9fe/metadata/instance?api-version=2021-02-01" \
+           "http://2852039166/metadata/instance?api-version=2021-02-01"; do
+  alt_body=/tmp/.gpa_c7_$$.body
```
srikrishnaveturi: Bug: Symlink race 

This harness runs as root (sudo ./run_all.sh). The temp file path is predictable, an attacker who knows the harness PID (or simply pre-creates files for a range of PIDs) can win a symlink race:


When curl -o "$alt_body" runs as root, it follows the symlink and overwrites /etc/shadow (or any other file). This is the exact attack class that scenarios E2 and F3 in DESIGN.md are designed to catch in GPA itself.

The fix is to use mktemp, which is already the pattern used 30 lines earlier in the same script for the B1a body file:


And then rm -f "$alt_body" in the loop cleanup stays the same.

pentest/linux/phase4b_local_rules/__pycache__/run.cpython-312.pyc
ZhidongPeng: good catch, it should been auto-ignored.

pentest/linux/phase4_rules_fuzz/url_diff.py
ZhidongPeng: thanks

ZhidongPeng: switched to `datetime.datetime.now(datetime.timezone.utc)`

pentest/linux/phase3_authn_authz/run.sh
ZhidongPeng: good catch.

srikrishnaveturi (approved):

PR #356, opened 2026-06-11
do not authenticate the elevated request if it designed to be skipped

ZhidongPeng:
Fix the issue if the customer's allowlist rules may block our telemetrydata or log upload requests, they are designed to be skipped from authenticating.

### DIFF
--- a/proxy_agent/src/proxy/proxy_server.rs
+++ b/proxy_agent/src/proxy/proxy_server.rs
@@ -475,53 +475,56 @@ impl ProxyServer {
         };
         http_connection_context.log(LoggerLevel::Trace, claim_details.to_string());
 
-        // authenticate the connection
-        let access_control_rules = match proxy_authorizer::get_access_control_rules(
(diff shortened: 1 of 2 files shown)


proxy_agent/src/proxy/proxy_server.rs
```diff
@@ -475,53 +475,56 @@ impl ProxyServer {
...
-        );
-        if result != AuthorizeResult::Ok {
-            // log to authorize failed connection summary
-            self.log_connection_summary(
-                &mut http_connection_context,
-                StatusCode::FORBIDDEN,
-                true,
-                "Authorize failed".to_string(),
+        // do not authenticate if the request is to be skipped
+        if !http_connection_context.should_skip_sig() {
```
srikrishnaveturi: Can we add tests for this behavior? Just to catch regressions in the future

srikrishnaveturi (approved):

PR #357, opened 2026-06-11
Introduce Canonical Request Model and with shadow canonical decision.

ZhidongPeng:
- Added CanonicalRequest model to load the proxied request for rule comparation
- Hoop up the authorization with shadow canonical decision, log only if canonical allowed is different than legacy allowed.
- Updated pen-tests


proxy_agent/src/proxy/proxy_connection.rs
```diff
@@ -282,8 +282,19 @@ impl HttpConnectionContext {
         hyper_client::should_skip_sig(&self.method, &self.url)
     }
 
+    /// Pre-canonical defense-in-depth guard. Returns `true` if the request
+    /// path contains a pattern commonly used to bypass prefix-based
+    /// authorization rules. See [`path_has_traversal`] for the exact set.
+    /// This is checked in `handle_new_http_request` before any rule lookup
+    /// so suspicious paths short-circuit to 404 without ever reaching the
```
srikrishnaveturi: Do we return 404? I think we return 403 forbidden code

ZhidongPeng: you are right, i need update the comment after I switched return 404 to 403.

proxy_agent/src/proxy/proxy_authorizer.rs
```diff
@@ -243,12 +244,50 @@ pub fn authorize(
...
+    // canonical matcher knows about. Comparing the wrapper output to
+    // the canonical matcher would produce phantom divergences. So we
+    // call the rules-only predicate twice in shadow/enforce — once
+    // through the Authorizer, once explicitly for the shadow
+    // comparison — accepting one extra matcher invocation per request
+    // only in non-production modes.
+    let mode = config::get_canonical_request_mode();
+    if mode != CanonicalMode::Off {
+        if let Some(rules) = access_control_rules.as_ref() {
+            let legacy_allowed = rules.is_allowed(logger, request_uri.clone(), claims.clone());
```
srikrishnaveturi, line 274: Is this different than previously created `legacy_result`?

srikrishnaveturi: Just read the comments, can discard this comment

ZhidongPeng: `legacy_result` returns the overall authorize result, including runAsElevated check for WS and rules.Mode audit mode check too.
`legacy_allowed` returns the rules-only check.
I have the comments mentioned above too
```
    // Why we re-evaluate `is_allowed` here instead of recovering it
    // from `legacy_result`: the wrapper above mixes in audit-mode
    // softening and `runAsElevated` gating, neither of which the
    // canonical matcher knows about. Comparing the wrapper output to
    // the canonical matcher would produce phantom divergences. So we
    // call the rules-only predicate twice in shadow/enforce
```

srikrishnaveturi (approved):

PR #366, opened 2026-06-25
Optimize telemetry event flow to by-pass local disk dependency

ZhidongPeng:
- `DirectSendConfig` defined in event_logger.rs and carries the execution_mode, event_name, optional version, and a CommonState handle (used to notify the sender).
- A background loop in `event_logger::start_with_direct_send`, with `DirectSendConfig`, wakes every 60 seconds and drains EVENT_QUEUE, each event is first attempted on the direct in-memory send path `event_sender::try_enqueue_generic_event`, which converts it and pushes it straight into the EventSender's TELEMETRY_EVENT_QUEUE. If at least one event was enqueued directly, the loop calls `common_state.notify_telemetry_event()` to wake the sender. Only events that cannot be enqueued directly fall back to disk.
- `report_extension_status_event` tries the direct path first via `event_sender::try_enqueue_extension_event` and, on success, notifies the sender and returns; otherwise it falls back to disk.
- Added TelemetryEventFlow.md file under doc folder for more design details.

### DIFF
--- a/proxy_agent/src/provision.rs
+++ b/proxy_agent/src/provision.rs
@@ -335,14 +335,22 @@ pub async fn start_event_threads(event_threads_shared_state: EventThreadsSharedS
     // those tasks starts to run after provision finished or provision timedout
     set_resource_limits();
 
-    let cloned_agent_status_shared_state =
-        event_threads_shared_state.agent_status_shared_state.clone();
     tokio::spawn({
-        async {
+        let cloned_agent_status_shared_state =
+            event_threads_shared_state.agent_status_shared_state.clone();
+        let direct_send_config = event_logger::DirectSendConfig::new(
+            "ProxyAgent".to_string(),
+            "MicrosoftAzureGuestProxyAgent".to_string(),
+            None,
+            event_threads_shared_state.common_state.clone(),
+        );
+
+        async move {
             event_logger::start(
                 config::get_events_dir(),
                 Duration::default(),
                 config::get_max_event_file_count(),
+                Some(direct_send_config),
                 move |status: String| {
                     let cloned_agent_status_shared_state = cloned_agent_status_shared_state.clone();
                     async move {
--- a/proxy_agent/src/provision.rs
+++ b/proxy_agent/src/provision.rs
@@ -346,11 +346,11 @@ pub async fn start_event_threads(event_threads_shared_state: EventThreadsSharedS
         );
 
         async move {
-            event_logger::start(
+            event_logger::start_with_direct_send(
                 config::get_events_dir(),
                 Duration::default(),
                 config::get_max_event_file_count(),
-                Some(direct_send_config),
+                direct_send_config,
                 move |status: String| {
                     let cloned_agent_status_shared_state = cloned_agent_status_shared_state.clone();
                     async move {
--- a/proxy_agent_shared/src/telemetry/event_logger.rs
+++ b/proxy_agent_shared/src/telemetry/event_logger.rs
@@ -47,37 +47,148 @@ impl DirectSendConfig {
     }
 }
 
+/// Start the telemetry event logger loop in disk-only mode.
+///
+/// Events are buffered to disk and later picked up by the event reader. This is
+/// used by callers (such as the extension) that do not have a direct in-memory
+/// telemetry path. It deliberately does not reference the `event_sender` direct
+/// path so that the direct-send machinery is not linked into binaries that only
+/// use this entry point.
 pub async fn start<F, Fut>(
     event_dir: PathBuf,
-    mut interval: Duration,
+    interval: Duration,
     max_event_file_count: usize,
-    direct_send_config: Option<DirectSendConfig>,
     set_status_fn: F,
 ) where
     F: Fn(String) -> Fut,
     Fut: std::future::Future<Output = ()>,
 {
+    // Disk-only mode: no events are sent directly, so every event falls through
+    // to be buffered on disk.
+    run_event_loop(
+        event_dir,
+        interval,
+        max_event_file_count,
+        set_status_fn,
+        |events| async move { events },
+    )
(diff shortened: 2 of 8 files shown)


proxy_agent_shared/src/telemetry/event_logger.rs
```diff
@@ -73,11 +108,49 @@ pub async fn start<F, Fut>(
...
         }
 
+        // Try to send the events directly via the in-memory telemetry queue
+        // first, so VMs without disk write permission can still report
+        // telemetry. Any events that cannot be enqueued directly (the direct
+        // path is None or the queue is full/closed) fall back to being
+        // buffered on disk below.
+        let events: Vec<Event> = match &direct_send_config {
+            Some(config) => {
+                let event_count = events.len();
```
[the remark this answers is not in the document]
ZhidongPeng: added more unit tests for try_direct_send_events(events: Vec<Event>, config: &DirectSendConfig)

proxy_agent/src/provision.rs
```diff
@@ -335,14 +335,22 @@ pub async fn start_event_threads(event_threads_shared_state: EventThreadsSharedS
...
-        event_threads_shared_state.agent_status_shared_state.clone();
     tokio::spawn({
-        async {
-            event_logger::start(
+        let cloned_agent_status_shared_state =
+            event_threads_shared_state.agent_status_shared_state.clone();
+        let direct_send_config = event_logger::DirectSendConfig::new(
+            "ProxyAgent".to_string(),
+            "MicrosoftAzureGuestProxyAgent".to_string(),
+            None,
```
srikrishnaveturi, line 344: in the disk path, are we also passing None? I think we're passing a variable instead of a constant None

proxy_agent_shared/src/telemetry/event_logger.rs
```diff
@@ -18,31 +19,176 @@ static EVENT_QUEUE: Lazy<ConcurrentQueue<Event>> =
...
+    interval: Duration,
+    max_event_file_count: usize,
+    direct_send_config: DirectSendConfig,
+    set_status_fn: F,
+) where
+    F: Fn(String) -> Fut,
+    Fut: std::future::Future<Output = ()>,
+{
+    if DIRECT_SEND_CONFIG.set(direct_send_config.clone()).is_err() {
+        let message = "DirectSendConfig is already set, cannot set it again.";
```
srikrishnaveturi, line 95: if we cannot set it again, the run_event_loop call below this should still use DIRECT_SEND_CONFIG right?

proxy_agent_shared/src/telemetry/event_reader.rs
```diff
@@ -305,22 +313,16 @@ impl EventReader {
...
-                let telemetry_event = TelemetryEvent::ExtensionEvent(telemetry_event);
-                event_sender::enqueue_event(telemetry_event);
-                if let Err(e) = self.common_state.notify_telemetry_event().await {
-                    logger_manager::write_warn(format!(
-                        "Failed to notify telemetry event with error: {e}"
-                    ));
+                if let Err(e) =
+                    event_sender::try_enqueue_extension_event(&event, self.execution_mode.clone())
+                {
+                    logger_manager::write_warn(format!("Failed to enqueue extension event: {}", e));
```
srikrishnaveturi, line 323: Are we okay with the file being cleaned up below `Self::clean_file(file);` even if the file enqueue failed?

proxy_agent_shared/src/telemetry/event_logger.rs
```diff
@@ -173,6 +325,27 @@ pub async fn report_extension_status_event(
...
+        // If the queue is full/closed, fall back to buffering the event on disk.
+        if crate::telemetry::event_sender::try_enqueue_extension_event(
+            &event,
+            config.execution_mode.clone(),
+        )
+        .is_ok()
+        {
+            if let Err(e) = config.common_state.notify_telemetry_event().await {
+                logger_manager::write_warn(format!(
+                        "report_extension_status_event: failed to notify telemetry event with error: {e}"
```
srikrishnaveturi, line 342: Can we have tests for this branch of code as well?

proxy_agent/src/provision.rs
ZhidongPeng: None here is for version, we using the current GPA fileversion.

ZhidongPeng: In `DirectSendConfig`, we do not need disk path, just need how to convert the event type into TelemetryGenericLog

proxy_agent_shared/src/telemetry/event_logger.rs
ZhidongPeng: this is ensure we set `DIRECT_SEND_CONFIG` once and for error handling only.

proxy_agent_shared/src/telemetry/event_reader.rs
ZhidongPeng: yeah, it is the existing logic for event_reader module, if could not enqueue, log-warn and remove the file. In the old code, the log-warn was in the enqueue_event() and never return error.
I touched this piece to reuse the common functions.

proxy_agent/src/provision.rs
ZhidongPeng: I see, you may mean if the disk path, the caller still call `start`, not the new `start_with_direct_send`, like our GPA VMExtension, it will write the events into VM Extension telemetry events folder.
This is ensuring no breaking change to existing callers.

proxy_agent_shared/src/telemetry/event_logger.rs
ZhidongPeng: Added tests `try_direct_send_events_notify_failed`

srikrishnaveturi (approved):

PR #372, opened 2026-07-07
Fix Linux extension exits 0 without writing status file on enable

srikrishnaveturi:
Three related defects caused the Linux GPA extension to exit 0 without writing a status file on enable, making waagent read back its own pre-created placeholder (" \n"), fail to parse it as JSON, and report VMExtensionProvisioningError to CRP.

Fix 1 -- proxy_agent_shared/src/service.rs:
query_service_executable_path now returns Result<PathBuf> instead of silently swallowing Err and returning an empty PathBuf. All callers are updated.

Fix 2 -- proxy_agent_extension/src/handler_main.rs:
enable_handler on Linux now always writes a Transitioning status for the current seq_no before exiting, regardless of whether the seq_no changed. Previously the write was guarded by should_report_status, so on VM restart with the same seq_no no status was written and waagent read back its own placeholder.

Fix 3 -- proxy_agent_extension/src/common.rs:
On Linux, get_proxy_agent_service_path() now returns the well-known fixed path (/usr/sbin/azure-proxy-agent) directly instead of querying systemctl show --property=ExecStart. The systemd unit-file cache is not hot right after install, causing ExecStart to return empty, which previously triggered a false VersionMismatch and a spurious re-install on every enable invocation.


proxy_agent_extension/src/handler_main.rs
```diff
@@ -228,14 +228,14 @@ fn uninstall_handler() {
...
-            if should_report_status {
-                common::report_status_enable_command(
-                    status_folder.to_path_buf(),
-                    &config_seq_no,
-                    None,
-                );
-            }
+        Ok(_should_report_status) => {
+            // Always write a Transitioning status for the current seq_no regardless
+            // of whether the seq_no changed.  waagent pre-allocates the status file
```
ZhidongPeng: This is not OK if the seq_no not changed at this enable command. VMAgent always re-run the enable command for the all the VM Extensions.

ZhidongPeng (changes requested):

proxy_agent_extension/src/handler_main.rs
srikrishnaveturi: Makes sense, reverted

ZhidongPeng (approved):

PR #375, opened 2026-07-16
Add evt_listener into windows_events module.

ZhidongPeng:
- Renamed `etw` mod to `windows_events` as the current sub mods are not actual event tracing library
- Added `models.rs` file under `windows_events` to capture the data models for the events
- Renamed etw sub mods with prefix `evt` instead of etw as they are calling classic windows event log library, `etw_reader` -> `evt_query`, `etw_writer` -> `evt_writer`
- Added `evt_listener` to subscribe windows event logs and tested together with `evt_writer` e2e.


proxy_agent_shared/src/windows_events/evt_query.rs
```diff
@@ -0,0 +1,228 @@
...
+/// Builds the `EvtQuery` XPath that selects entries whose `Provider @Name`
+/// equals `source_name` and that fall within the optional inclusive
+/// `[start_time, end_time]` range, so the Event Log service performs the
+/// filtering instead of the iterator.
+fn build_query(
+    source_name: &str,
+    start_time: Option<DateTime<Utc>>,
+    end_time: Option<DateTime<Utc>>,
+) -> String {
+    let mut predicates = vec![format!("Provider[@Name={}]", xpath_literal(source_name))];
```
bingyuyin, line 148: will it resolved similar as below? like a list of Name vs EnventIdLevelMapping.


'EventSources': [{
                                          'Name': '.NET Runtime','EventIdLevelMapping': ['1026: Error']
                                          },
                                        {
                                          'Name': 'MsiInstaller','EventIdLevelMapping': ['1040: Important', '1035: Important']
                                        },
                                        {
                                          'Name': 'Application Error','EventIdLevelMapping': ['1000: Error']
                                        },
                                          {
                                              'Name': 'User32', 'EventIdLevelMapping': ['1074: Important']
                                          }]

ZhidongPeng: no, it filters at windows api side and only the selected source_name events will be queried in our rust user code. hence it removed the matches in evt_query.rs file.

ZhidongPeng: Similarly in evt_listener.rs, I have `struct SourceFilter` which defines the source name and its eventIds, so less events will be returned into callback function.

ZhidongPeng: Also, there is no `EventIdLevelMapping` to particular LEVEL string, it uses the original level from event logs directly.

bingyuyin (approved):

PR #378, opened 2026-07-17
Add etw_listener to help capture ebpf-for-windows tracing events.

ZhidongPeng:


proxy_agent_shared/src/windows_events/etw_listener.rs
```diff
@@ -0,0 +1,1061 @@
...
+    pub fn add_provider(&mut self, provider_id: &str, max_level: u8) -> Result<()> {
+        if let Some(guid) = parse_guid_string(provider_id) {
+            self.providers.push(EtwProvider {
+                provider_id: guid,
+                level: max_level,
+                event_ids: Vec::new(),
+            });
+            return Ok(());
+        }
+        if let Some(guid) = resolve_provider_name(provider_id) {
```
bingyuyin: nit: name instead of guid

proxy_agent_shared/src/windows_events/etw_listener.rs
```diff
@@ -0,0 +1,1061 @@
...
+        }
+    }
+}
+
+/// Runs the full trace lifecycle: start session, enable providers, open and
+/// process the real-time trace, then clean up. Each decoded event is delivered
+/// to `handler`, whose address is passed to ETW via the log file `Context` and
+/// recovered in the callback from `EVENT_RECORD.UserContext`. `handler` must
+/// outlive this call (it does: this function blocks until the trace stops).
+fn run_trace(
```
bingyuyin, line 251: make is async function instead of doing a thread.sleep per loop

bingyuyin (dismissed):
approved with minor comments

proxy_agent_shared/src/windows_events/etw_listener.rs
ZhidongPeng: cannot make it async function as whose sole job is to block on `ProcessTrace`. Blocking a thread that exists only for that purpose is fine and idiomatic.

ZhidongPeng: I fixed the caller to use `std::thread::spawn` instead of `tokio::spawn` with more detailed comments.

proxy_agent_shared/src/windows_events/etw_listener.rs
ZhidongPeng: it is actually a provider id in guid format, let me rename it to `provider_guid`.

bingyuyin (dismissed):

proxy_agent_shared/src/windows_events/etw_listener.rs
```diff
@@ -0,0 +1,1076 @@
...
+        let session_name = self.session_name.clone();
+
+        // `run_trace` blocks on `ProcessTrace` for the lifetime of the session,
+        // so run it on a dedicated OS thread.
+        // Note: do not use tokio::spawn because the async block runs on a tokio worker thread.
+        // Since run_trace blocks without ever yielding, it monopolizes that worker thread for the whole session.
+        // Tokio's worker pool is small (roughly one per core), so, it permanently removes a worker from the pool,
+        //  starving other async tasks and risking degradation/deadlock.
+        // This is exactly the "don't block the async runtime" anti-pattern.
+        std::thread::spawn(move || {
```
srikrishnaveturi: The thread this creates is unnamed - can we add a name in case we need to track this thread?

proxy_agent_shared/src/telemetry/event_sender.rs
```diff
@@ -164,6 +166,9 @@ impl EventSender {
                     logger_manager::write_warn(format!(
                         "Failed to send telemetry data to host with error: {e}"
                     ));
+                    // Dev debug log the telemetry data to help with troubleshooting
+                    logger_manager::write_log(LoggerLevel::Trace, xml_data.clone());
```
srikrishnaveturi: Does this whole xml entry need to be logged on each retry? Can we log this entry outside the look and just print retry numbers inside the loop?

srikrishnaveturi (approved):

PR #386, opened 2026-07-30
read and deserialize json UTF-8, UTF-16, UTF-32, BOM, LE/BE support

bingyuyin:
read and deserialize json file support all encoding types regarding UTF-8, UTF-16, UTF-32, BOM, LE/BE


proxy_agent_shared/src/misc_helpers.rs
```diff
@@ -245,26 +246,134 @@ where
...
+/// Detects the text encoding of `bytes` and returns it as
+/// (code unit width in bytes, big endian, BOM length in bytes).
+/// width: 1 for UTF-8, 2 for UTF-16, 4 for UTF-32
+/// big_endian: true for BE, false for LE
+/// bom length length of bom
+///
+/// Wider BOMs must be tested first: the UTF-32LE BOM (FF FE 00 00) starts with
+/// the UTF-16LE BOM (FF FE), so a shortest-first scan would mis-detect a
+/// UTF-32LE file as UTF-16LE.
+fn detect_json_encoding(bytes: &[u8]) -> (usize, bool, usize) {
```
ZhidongPeng: can we wrap these return values into a struct with clear field names?

proxy_agent_shared/src/misc_helpers.rs
```diff
@@ -245,26 +246,134 @@ where
...
+    if bytes.starts_with(&[0x00, 0x00, 0xFE, 0xFF]) {
```
ZhidongPeng, line 309: have you checked if the input `bytes` only has 1 or 2 byte? unit tests for such edge case coverages?

proxy_agent_shared/src/misc_helpers.rs
```diff
@@ -245,26 +246,134 @@ where
...
+        } else {
+            (1, false, 0) // anything else, including plain ASCII / UTF-8
+        }
+    }
+}
+
+/// Decodes `bytes` into UTF-8 text using the encoding detected by
+/// [`detect_json_encoding`], skipping the BOM when present.
+/// UTF-8 input is borrowed as-is, so the common case does not allocate.
+fn decode_json_text<'a>(bytes: &'a [u8], file_path: &Path) -> Result<Cow<'a, str>> {
```
ZhidongPeng: this fn has nothing to do with json text, same as `detect_json_encoding`, suggest having more generic function names.

proxy_agent_shared/src/misc_helpers.rs
```diff
@@ -245,26 +246,134 @@ where
...
+    let decode_error = |detail: String| Error::DecodeFile(path_to_string(file_path), detail);
+
+    let (code_unit_len, big_endian, bom_len) = detect_json_encoding(bytes);
+    let payload = &bytes[bom_len..];
+
+    match code_unit_len {
+        1 => std::str::from_utf8(payload)
```
ZhidongPeng: please use enum instead of the magic numbers

proxy_agent_shared/src/misc_helpers.rs
```diff
@@ -245,26 +246,134 @@ where
...
+            (1, false, 0) // anything else, including plain ASCII / UTF-8
+        }
+    }
+}
+
+/// Decodes `bytes` into UTF-8 text using the encoding detected by
+/// [`detect_json_encoding`], skipping the BOM when present.
+/// UTF-8 input is borrowed as-is, so the common case does not allocate.
+fn decode_json_text<'a>(bytes: &'a [u8], file_path: &Path) -> Result<Cow<'a, str>> {
+    let decode_error = |detail: String| Error::DecodeFile(path_to_string(file_path), detail);
```
ZhidongPeng: Prefer `Err::DcodeBytes` and get rid of `file_path` as the input content is `bytes`

proxy_agent_shared/src/misc_helpers.rs
```diff
@@ -644,6 +753,65 @@ mod tests {
...
+        #[derive(Serialize, Deserialize, PartialEq, Debug)]
+        struct EncodingTestStruct {
+            name: String,
+            code: i32,
+            message: String,
+            enabled: bool,
+        }
+
+        let test_data_dir = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+            .join("test_data")
```
ZhidongPeng: please help remove test_data files, lets generate and remove those test files at before and end of test executions

ZhidongPeng (changes requested):

proxy_agent_shared/src/misc_helpers.rs
bingyuyin: for those cases like bytes only has 1 or 2 byte or invalid format cannot detect the encoding type, it will fallback to default utf 8 without bom, no need further protection check, the bytes.get(i) is safe boundary check return option

proxy_agent_shared/src/misc_helpers.rs
bingyuyin: Done

proxy_agent_shared/src/misc_helpers.rs
bingyuyin: added a new unit test to test, the empty, 1 or 2 byte that are invalid will fallback to default utf8 with no bom

proxy_agent_shared/src/misc_helpers.rs
bingyuyin: done

proxy_agent_shared/src/misc_helpers.rs
bingyuyin: done

proxy_agent_shared/src/misc_helpers.rs
bingyuyin: get rid of the file_path as parameter, to make the caller map the error with file path

proxy_agent_shared/src/misc_helpers.rs
bingyuyin: renamed function

ZhidongPeng (approved):

PR #388, opened 2026-08-27
GPA to support IPv6 dual-stack address

ZhidongPeng:
**Context**: Some of the modern http client library implicitly use IPv6 dual-stack mode for IPv4 host ipaddress, while our current GPA/eBPF program only hook up cgroup/connect4 event. So, in enforce mode, such modern http requests will goto host endpoint directly without captured/proxied by GPA/eBPF, then these requests rejected by host with 401.

**Fix**: 
- GPA to hook up /attaches connect6 event
- The connect6 programs recognize the ::ffff:0:0/96 prefix, extract the low-order IPv4 address, and look it up in the existing IPv4 policy map. A matched connection is redirected to IPv4-mapped loopback while remaining usable by the originating dual-stack socket.
- Add addressFamily to the ProxySummary telemetry event

Native IPv6 still in unsupported until we published the well-known host IPv6 address:
- Native IPv6 addresses do not match the IPv4 policy map and pass through unchanged
- No new native IPv6 endpoint configuration is introduced in this stage


e2etest/GuestProxyAgentTest/TestCases/IMDSPingTestCase.cs
```diff
@@ -8,17 +8,20 @@ namespace GuestProxyAgentTest.TestCases
...
         }
 
         private bool ImdsSecureChannelEnabled { get; set; }
+        private bool Ipv6DualStackSupported { get; set; }
 
         public override async Task StartAsync(TestCaseExecutionContext context)
         {
             List<(string, string)> parameterList = new List<(string, string)>();
             parameterList.Add(("imdsSecureChannelEnabled", ImdsSecureChannelEnabled.ToString()));
+            parameterList.Add(("ipv6DualStackSupported", Ipv6DualStackSupported.ToString()));
```
srikrishnaveturi, line 24: Is this parameter being added for both windows and linux? Did we make changes to the windows script to accept this param?

ZhidongPeng: this parameter added for both windows and linux; while windows support it by default as it convert ipv6 dual stack to ipv4 implicitly. so, I do not check this new parameter.

ZhidongPeng: more captured here: https://github.com/microsoft/ebpf-for-windows/issues/5536

srikrishnaveturi (approved):

PR #389, opened 2026-08-28
Skip connections bind to local IPAddress

ZhidongPeng:
**Context**:
Client http request bind to local ip (ex 10.0.0.4) when communicating to the host endpoints, for example:
- uses [ServicePoint.BindIPEndPointDelegate](https://learn.microsoft.com/en-us/dotnet/api/system.net.servicepoint.bindipendpointdelegate?view=netframework-4.0) or
- uses `curl --interface 10.0.0.4` [curl(1) - Linux manual page](https://www.man7.org/linux/man-pages/man1/curl.1.html). 
 
GPA ProxyServer listens to loopback/localhost IP, but OS networking kernel, not eBPF, could NOT redirect such requests to loopback/localhost IP address. The OS rejects this because:
- **Loopback is a closed loop**: Traffic on 127.0.0.0/8 is expected to originate and terminate on the loopback interface. A packet with source 10.0.0.4 arriving on loopback violates this assumption.
- **Reverse path validation**: The kernel (Linux rp_filter, Windows strong host model) checks that the source IP is reachable via the interface the packet arrives on. 10.0.0.4 is not reachable via loopback, so the packet is dropped.
- **Route selection**: The kernel picks the outgoing interface based on the destination. For 127.0.0.1, it selects loopback. But the bound source IP doesn't belong to loopback, causing a "network unreachable" or silent drop.

**Changes**:
- Add ebpf map `config_map` to accept settings `GPA_CONFIG_LOCAL_IP_BIND_MONITOR_ONLY`
- Add ebpf map `audit_only_map` to cache connections bind to local IP address
- if `GPA_CONFIG_LOCAL_IP_BIND_MONITOR_ONLY` set, skip the redirect if the connections bind to local IP Address 
- Start `poll_audit_only` task to read/pop the elem from ebpf map `audit_only_map` and emit telemetry event every few seconds.

**Known issue**:
- As the `poll_audit_only` task to read/pop every few seconds, the process details may not available if the process exits quickly

### DIFF
--- a/proxy_agent/src/redirector.rs
+++ b/proxy_agent/src/redirector.rs
@@ -53,15 +53,12 @@ use crate::common::helpers;
 use crate::common::result::Result;
 use crate::common::{config, logger};
 use crate::provision;
-use crate::shared_state::access_control_wrapper::AccessControlSharedState;
+use crate::proxy::Claims;
 use crate::shared_state::agent_status_wrapper::{AgentStatusModule, AgentStatusSharedState};
-use crate::shared_state::connection_summary_wrapper::ConnectionSummarySharedState;
-use crate::shared_state::key_keeper_wrapper::KeyKeeperSharedState;
-use crate::shared_state::provision_wrapper::ProvisionSharedState;
+use crate::shared_state::proxy_server_wrapper::ProxyServerSharedState;
 use crate::shared_state::redirector_wrapper::RedirectorSharedState;
 use crate::shared_state::EventThreadsSharedState;
 use crate::shared_state::SharedState;
-use proxy_agent_shared::common_state::CommonState;
 use proxy_agent_shared::logger::LoggerLevel;
 use proxy_agent_shared::misc_helpers;
 use proxy_agent_shared::proxy_agent_aggregate_status::ModuleState;
@@ -109,28 +106,14 @@ impl AuditEntry {
 
 pub struct Redirector {
     local_port: u16,
-    redirector_shared_state: RedirectorSharedState,
-    key_keeper_shared_state: KeyKeeperSharedState,
-    agent_status_shared_state: AgentStatusSharedState,
-    cancellation_token: CancellationToken,
-    common_state: CommonState,
-    provision_shared_state: ProvisionSharedState,
-    access_control_shared_state: AccessControlSharedState,
-    connection_summary_shared_state: ConnectionSummarySharedState,
+    shared_state: SharedState,
 }
 
 impl Redirector {
     pub fn new(local_port: u16, shared_state: &SharedState) -> Self {
         Redirector {
             local_port,
-            cancellation_token: shared_state.get_cancellation_token(),
-            key_keeper_shared_state: shared_state.get_key_keeper_shared_state(),
-            common_state: shared_state.get_common_state(),
-            provision_shared_state: shared_state.get_provision_shared_state(),
-            agent_status_shared_state: shared_state.get_agent_status_shared_state(),
-            redirector_shared_state: shared_state.get_redirector_shared_state(),
-            access_control_shared_state: shared_state.get_access_control_shared_state(),
-            connection_summary_shared_state: shared_state.get_connection_summary_shared_state(),
+            shared_state: shared_state.clone(),
         }
     }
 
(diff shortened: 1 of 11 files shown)


proxy_agent/src/redirector/shared_ebpf.rs
```diff
@@ -140,6 +141,45 @@ pub struct sock_addr_audit_entry {
     pub address_family: u32,
     pub reserved: u32,
 }
+
+#[repr(C)]
+#[derive(Clone, Copy, Debug)]
+pub struct audit_only_event {
```
ZhidongPeng, line 147: it does not match gpa_audit_event.h
```
// Audit-only ring-buffer record. The kernel timestamp is monotonic; user mode
// adds UTC at receipt because eBPF does not expose a UTC clock on all platforms.
struct gpa_audit_only_event
{
    __u64 kernel_timestamp_ns;
    __u32 local_ipv4;
    struct gpa_audit_event audit;
};
```

proxy_agent/src/redirector.rs
```diff
@@ -106,6 +103,13 @@ pub struct AuditEntry {
     pub address_family: AddressFamily,
 }
 
+pub struct AuditOnlyRecord {
+    pub entry: AuditEntry,
+    pub kernel_timestamp_ns: u64,
```
ZhidongPeng, line 108: we need to use this `kernel_timestamp_ns` to tell this pid is a valid one or not.

proxy_agent/src/redirector.rs
```diff
@@ -106,6 +103,13 @@ pub struct AuditEntry {
...
+    pub timestamp_utc_ns: i128,
+    pub local_ipv4: u32,
```
ZhidongPeng, line 110: we may need to have ipv6?

proxy_agent/src/redirector.rs
```diff
@@ -294,6 +301,64 @@ impl Redirector {
...
+                            let message = match Claims::from_audit_entry(
+                                &entry,
+                                std::net::IpAddr::V4(std::net::Ipv4Addr::UNSPECIFIED), // not used for audit-only, so just use unspecified
+                                0, // not used for audit-only, so just use 0
+                                proxy_server_shared_state.clone(),
+                            )
+                            .await
+                            {
+                                Ok(claims) => format!(
+                                    "eBPF audit-only connection: timestampUtcNs={}, kernelTimestampNs={}, localIp={}, userName={}, processId={}, processName={}, processFullPath={}, processCmdLine={}, runAsElevated={}, destination={}:{}",
```
ZhidongPeng: message prefix is unnecessary.

ZhidongPeng: `timestampUtcNs` and `kernelTimestampNs` are not human readable

proxy_agent/src/redirector.rs
```diff
@@ -294,6 +301,64 @@ impl Redirector {
...
+        tokio::select! {
+            _ = cancellation_token.cancelled() => return,
+            record = receiver.recv() => {
+                let Some(record) = record else { return; };
+                let entry = record.entry;
+                            let destination_ip = entry.destination_ipv4_addr();
+                            let destination_port = entry.destination_port_in_host_byte_order();
+                            let message = match Claims::from_audit_entry(
+                                &entry,
+                                std::net::IpAddr::V4(std::net::Ipv4Addr::UNSPECIFIED), // not used for audit-only, so just use unspecified
```
ZhidongPeng: please indicate why they are not used, are they included into the telemetry message?

ZhidongPeng: client_ip is needed, 
local_port is not needed/used.

proxy_agent/src/redirector.rs
```diff
@@ -294,6 +301,64 @@ impl Redirector {
...
+                                    Ipv4Addr::from_bits(record.local_ipv4.to_be()),
+                                    entry.logon_id,
+                                    entry.process_id,
+                                    destination_ip,
+                                    destination_port,
+                                ),
+                            };
+                            event_logger::write_event(
+                                LoggerLevel::Warn,
+                                message,
```
ZhidongPeng: how to query or filter the message? let's use json format for the message.

